PR #1, opened 2017-11-14
Import script access controls

smithjp:


lib/tasks/migrate.rake
```diff
@@ -168,6 +175,26 @@ namespace :cdr do
       publisher = mods_version.xpath('mods:originInfo//mods:publisher',MigrationConstants::NS).text
       language = mods_version.xpath('mods:language//mods:languageTerm',MigrationConstants::NS).text
 
+      if rdf_version.to_s.match(/metadata-patron/)
+        patron = rdf_version.xpath("rdf:Description/*[local-name() = 'metadata-patron']", MigrationConstants::NS).text
```
bbpennel: What objects are we ingesting which use this setting?  If i'm understanding, it looks like metadata only access is from the CDR is being translated into any authenticated user in hyc?  Or is it more nuanced than that?

Spoke with Julie a bit about this, I think the correct behavior for metadata patron access would be for the work to be visible to public, but the files to be private

lib/tasks/migrate.rake
```diff
@@ -168,6 +175,26 @@ namespace :cdr do
...
+      elsif rdf_version.to_s.match(/embargo-until/)
+        embargo_release_date = Date.parse rdf_version.xpath("rdf:Description/*[local-name() = 'embargo-until']", MigrationConstants::NS).text
+        visibility = visibility_during_embargo
+      elsif rdf_version.to_s.match(/isPublished/)
+        published = rdf_version.xpath("rdf:Description/*[local-name() = 'isPublished']", MigrationConstants::NS).text
+        if published == 'no'
+          visibility = Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PRIVATE
+        end
+      elsif rdf_version.to_s.match(/inheritPermissions/)
+        published = rdf_version.xpath("rdf:Description/*[local-name() = 'inheritPermissions']", MigrationConstants::NS).text
```
bbpennel: might want to use a different variable name here, repeating `published` threw me off at first

lib/tasks/migrate.rake
bbpennel: you may also want a case for regular patron roles assigned to `authenticated` and `public` roles.  For example, this paper is only accessible to authenticated users:
https://cdr.lib.unc.edu/record/uuid:8a8abab5-3e17-498d-8084-9a2c15e477da

lib/tasks/migrate.rake
```diff
@@ -88,8 +88,10 @@ namespace :cdr do
       file_set = FileSet.new
       file_set.title = resource[:resource][:title]
       actor = ::Hyrax::Actors::FileSetActor.new(file_set, User.find_by_email(DEPOSITOR_EMAIL))
+      actor.create_metadata(resource)
       actor.create_content(File.open(f))
       actor.attach_to_work(parent)
+      actor.file_set.permissions_attributes = parent.permissions.map(&:to_hash)
```
bbpennel: Am I correct that this is telling files to inherit the permissions of their parent?  I think this is just a lack of samvera familiarity on my part, but is the parent here the Work, or a child of a work?

Asking because we do have use cases for files to have separate permissions from the work they are owned by (cover pages for honor's theses for example, and maybe article xml files from biomed)

lib/tasks/migration/tmp/22f2d84b-8be9-410b-be4d-afd678870a12/uuid:22f2d84b-8be9-410b-be4d-afd678870a12-object.xml
```diff
@@ -43,7 +43,7 @@
     <foxml:datastreamVersion ID="RELS-EXT.1" LABEL="Fedora Object-to-Object Relationship Metadata" CREATED="2017-06-15T19:22:24.429Z" MIMETYPE="text/xml" SIZE="4020">
       <foxml:contentDigest TYPE="MD5" DIGEST="ae2c7a001ad9bb5467e40194102f638e"/>
       <foxml:xmlContent>
-        <rdf:RDF xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:dcterms="http://purl.org/dc/terms/" xmlns:ns5="http://cdr.unc.edu/definitions/1.0/base-model.xml#" xmlns:ns6="info:fedora/fedora-system:def/model#" xmlns:owl="http://www.w3.org/2002/07/owl#" xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#" xmlns:rdfs="http://www.w3.org/2000/01/rdf-schema#" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+        <rdf:RDF xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:dcterms="http://purl.org/dc/terms/" xmlns:ns5="http://cdr.unc.edu/definitions/1.0/base-model.xml#" xmlns:ns6="info:fedora/fedora-system:def/model#" xmlns:owl="http://www.w3.org/2002/07/owl#" xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#" xmlns:rdfs="http://www.w3.org/2000/01/rdf-schema#" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:cdr-role="something">
```
bbpennel: In case you want the namespace for roles, it is `http://cdr.unc.edu/definitions/roles#`

PR #2, opened 2017-11-27
Add worktype tests

smithjp:
@bbpennel @lfarrell


Gemfile
```diff
@@ -40,9 +40,19 @@ gem 'jbuilder', '~> 2.5'
 # Use Capistrano for deployment
 # gem 'capistrano-rails', group: :development
 
+# Windows does not include zoneinfo files, so bundle the tzinfo-data gem
+gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby]
+
+gem 'rsolr', '>= 1.0'
```
bbpennel: Generally with the new gems here, I wonder if we should default to specifying gems to just allow variation on patch releases?

app/forms/hyrax/work_form.rb
```diff
@@ -22,7 +22,11 @@ def self.model_attributes(_)
 
       single_value_fields.each do |field|
         if attrs[field]
-          attrs[field] = Array(attrs[field])
+          if attrs[field].blank?
```
bbpennel: this seems to be a recurring theme.  I wonder if it would make sense to factor this out into a common component?  If that is impractical or not worthwhile, that's okay too, it just stands out after the 5th repetition.

spec/presenters/hyrax/masters_paper_presenter_spec.rb
```diff
@@ -3,7 +3,120 @@
...
+        allow(Hyrax::Renderers::AttributeRenderer).to receive(:new).with(:author_degree_granted, ['MSIS'], {}).and_return(renderer)
+      end
+
+      it "calls the AttributeRenderer" do
+        expect(renderer).to receive(:render)
+        presenter.attribute_to_html(:author_degree_granted)
+      end
+    end
+
+    context "with a field that doesn't exist" do
```
bbpennel: This is more a question than anything, but is it normal practice to test that a log statement happens?  I can see the value in error cases, just seems like it could be a lot tests

spec/services/author_status_service_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+RSpec.describe Hyrax::AuthorStatusService do
+  before do
+    # Configure QA to use fixtures
+    qa_fixtures = { local_path: File.expand_path('spec/fixtures/authorities') }
+    allow(Qa::Authorities::Local).to receive(:config).and_return(qa_fixtures)
+  end
+  let(:service) { described_class }
+
+  # describe "#select_active_options" do
```
bbpennel: Are these commented out cases going to be used in the future?

spec/presenters/hyrax/masters_paper_presenter_spec.rb
smithjp: The hyrax gem tests included a similar type of test for error cases, but I think it is odd to have the error case check for a log statement.  I have replaced the check for a log statement by checking that the render does not occur.

Gemfile
smithjp: I have added restrictions to all gems listed since it has been a problem in the past.

PR #4, opened 2017-12-20
Honors thesis workflow

smithjp:
Added workflow method to allow review permissions to be assigned at a work level.


config/workflows/honors_thesis_deposit_workflow.json
```diff
@@ -0,0 +1,76 @@
...
+                    ]
+                }, {
+                    "name": "request_changes",
+                    "from_states": [{"names": ["deposited", "pending_review"], "roles": ["approving"]}],
+                    "transition_to": "changes_required",
+                    "notifications": [
+                        {
+                            "notification_type": "email",
+                            "name": "Hyrax::Workflow::ChangesRequiredNotification",
+                            "to": ["approving"]
```
bbpennel: I'm not 100% sure I'm reading this correctly, but does this `request changes` send an email notification to the same people as `deposit`?  Should it be going to "depositing" user instead, if the reviewer is asking for changes?

Same question regarding the `approve` transition.

app/lib/status_list_service_helper.rb
```diff
@@ -0,0 +1,50 @@
+# Overriding workflow permissions methods
+Hyrax::Workflow::StatusListService.class_eval do
+
+  private
+
+    def query(actionable_roles)
```
bbpennel: I don't see any issues, but I'm not really familiar with the code being overridden where and in ability.rb.  Are both of these changes things that we are going to push upstream?  Asking since I would guess they would be in a much better place to review this than we are.

config/workflows/honors_thesis_deposit_workflow.json
smithjp: It does appear to send the email to the reviewer instead of the depositor, which is wrong.  I will change this here and in the default mediated workflow.

app/lib/status_list_service_helper.rb
```diff
@@ -0,0 +1,51 @@
...
```
lfarrell: As best as I can tell, everything looks good.

PR #5, opened 2017-12-22
Honors thesis work type

smithjp:


app/forms/hyrax/honors_thesis_form.rb
```diff
@@ -0,0 +1,23 @@
...
+module Hyrax
+  class HonorsThesisForm < ::SingleValueForm
+    self.model_class = ::HonorsThesis
+
+    class_attribute :single_value_fields
+
+    self.terms += [:abstract, :academic_concentration, :academic_department, :access, :advisor, :degree,
+                   :degree_granting_institution, :genre, :graduation_year, :honors_level, :note, :resource_type]
+    self.terms -= [:based_near, :contributor, :date_created, :description, :identifier, :publisher, :source]
+    self.required_fields += [:abstract, :academic_department, :advisor, :degree_granting_institution, :resource_type,
```
bbpennel: In the map spreadsheet, I see the following as {0,n} which I interpret as meaning they aren't required:
* abstract
* resource_type
* advisor
* department

Seems like license may be in the same boat if i'm interpreting things correctly, where `Use` from the spreadsheet is license in hyrax?

app/models/concerns/hyrax/basic_metadata.rb
```diff
@@ -0,0 +1,42 @@
...
+      property :contributor, predicate: ::RDF::Vocab::DC11.contributor
+      property :description, predicate: ::RDF::Vocab::DC11.description
+      property :keyword, predicate: ::RDF::Vocab::SCHEMA.keywords
+      # Used for a license
+      property :license, predicate: ::RDF::Vocab::DC11.rights
+
+      # This is for the rights statement
+      property :rights_statement, predicate: ::RDF::Vocab::EDM.rights
+      property :publisher, predicate: ::RDF::Vocab::DC11.publisher
+      property :date_created, predicate: ::RDF::Vocab::DC.created
```
bbpennel: Legacy data will likely have date_issued, even if we aren't currently intending to let new papers submit them.  Will it still be able to be mapped if its not in the model here?

app/models/concerns/hyrax/basic_metadata.rb
```diff
@@ -0,0 +1,42 @@
...
+  # An optional model mixin to define some simple properties. This must be mixed
+  # after all other properties are defined because no other properties will
+  # be defined once  accepts_nested_attributes_for is called
+  module BasicMetadata
+    extend ActiveSupport::Concern
+
+    included do
+      property :label, predicate: ActiveFedora::RDF::Fcrepo::Model.downloadFilename, multiple: false
+
+      property :relative_path, predicate: ::RDF::URI.new('http://scholarsphere.psu.edu/ns#relativePath'), multiple: false
```
bbpennel: Are we okay with using scholarsphere properties?  Did anna have any suggestions for these predicates?  We could also create our own predicates, just feels a bit weird to me having their repo namespace in here if its not published.

app/models/honors_thesis.rb
```diff
@@ -0,0 +1,67 @@
...
+  end
+
+  property :note, predicate: ::RDF::Vocab::SKOS.note do |index|
+    index.as :stored_searchable
+  end
+
+  property :rights, predicate: ::RDF::Vocab::DC11.rights do |index|
+    index.as :stored_searchable
+  end
+
```
bbpennel: Wonder if `extent` needs to be mapped?

app/presenters/hyrax/honors_thesis_presenter.rb
```diff
@@ -0,0 +1,8 @@
+# Generated via
+#  `rails generate hyrax:work HonorsThesis`
+module Hyrax
+  class HonorsThesisPresenter < Hyrax::WorkShowPresenter
+    delegate :abstract, :academic_department, :academic_concentration, :access, :advisor, :degree,
```
bbpennel: Sorry for my ignorance, but i assume there are other default fields that get presented without having to list them here?  Like title and creator

spec/actors/hyrax/actors/honors_thesis_actor_spec.rb
```diff
@@ -0,0 +1,9 @@
+# Generated via
+#  `rails generate hyrax:work HonorsThesis`
+require 'rails_helper'
+
+RSpec.describe Hyrax::Actors::HonorsThesisActor do
+  it "has tests" do
+    skip "Add your tests here"
```
bbpennel: Do any tests need to be added here?

app/forms/hyrax/honors_thesis_form.rb
smithjp: I initially added everything Anna had marked as "should" in the MAP as required fields since the abstract field is required for google scholar and the department field is required for assigning a reviewer for honors theses.  I think that this need to be discussed further with Anna, Rebekah, and Julie, so I will update the code to reflect that they are not required fields.

app/models/concerns/hyrax/basic_metadata.rb
smithjp: This file defines the default metadata from the hyrax gem.  I had to override it to change the predicates for keyword and license properties, but I did not ask Anna about the rest.  I will ask her if any of the others need to be changed as well.

app/models/concerns/hyrax/basic_metadata.rb
smithjp: The date_issued field will be available for work types that have it defined as a custom field since it is not a default metadata property in hyrax.  I think for this work type Anna had marked it as "should not" in the MAP, so I did not include it in the model file.

app/models/honors_thesis.rb
smithjp: I think that was another property Anna had marked as "should not" in the MAP for honors theses works.

app/presenters/hyrax/honors_thesis_presenter.rb
smithjp: Yes, everything listed here is a custom metadata property.  The default metadata properties listed in the basic_metadata file are automatically delegated to the solr document.

spec/actors/hyrax/actors/honors_thesis_actor_spec.rb
smithjp: Sorry.  I forgot to remove this file after it was generated.  I didn't make any changes to the actor or controller files, so I shouldn't need to test anything here, assuming that the default functionality is being thoroughly tested in the gem.

app/models/concerns/hyrax/basic_metadata.rb
bbpennel: Okay, thanks.  I asked Anna about the two "should not" date fields.  Her answer was that new papers wouldn't add these fields, but ones being imported from the current repo would have the field, and that we would want to retain it.

app/models/honors_thesis.rb
bbpennel: I see it marked as "may" in the spreadsheet, she may have changed it at some point

smithjp: Ah, I see it.  She must have changed it after the ticket was marked as done.  I'll add it to the model.

app/models/concerns/hyrax/basic_metadata.rb
smithjp: Ok.  I'll add those fields to the model but not to the form and test a file import to make sure everything works.  That should allow us to retain the date_issued and date_created properties for old papers but not add them for new ones.

app/models/concerns/hyrax/basic_metadata.rb
smithjp: Anna said that the other default properties were ok, so I'm going to leave this as it is for now.  I also need to change the predicate for subject back to DC11.

PR #8, opened 2018-01-22
Dissertation work type

smithjp:


spec/features/create_work_spec.rb
```diff
@@ -6,28 +6,54 @@
...
+
+      click_link "Files" # switch tab
+      within "//span[@id=addfiles]" do
+        attach_file('files[]', File.join(Rails.root, '/spec/fixtures/files/test.txt'))
+      end
+      click_button 'Save'
+      expect(page).to have_content 'Your files are being processed by Hyrax'
+      visit '/dashboard/my/works/'
+      expect(page).to have_content 'Test Work work'
+      # within '.document-title' do
```
bbpennel: Do these commented out steps need to be kept?

spec/forms/hyrax/dissertation_form_spec.rb
```diff
@@ -9,62 +9,83 @@
...
           thumbnail_id: '789',
           keyword: ['derp'],
-          member_of_collection_ids: ['123456', 'abcdef']
+          member_of_collection_ids: ['123456', 'abcdef'],
+          abstract: ['an abstract'],
+          access: 'public',
+          date_issued: '2018-01-08',
+          degree: 'MSIS',
+          degree_granting_institution: 'UNC',
+          doi: 'hi.org',
```
bbpennel: `Handicap International / Humanity & Inclusion`, cool!

config/authorities/rights_statements.yml
```diff
@@ -35,3 +35,6 @@ terms:
 - id: http://rightsstatements.org/vocab/NKC/1.0/
   term: "No Known Copyright"
   active: true
+- id: http://www.europeana.eu/portal/rights/rr-r.html
+  term: "Old - Do Not Use"
```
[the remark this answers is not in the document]
lfarrell: This may only apply to me. I had a bunch of imported records with the old, broken rights statements. I could rebuild the VM and I think that would fix it.

app/models/solr_document.rb
```diff
@@ -72,6 +72,10 @@ def citation
     self[Solrizer.solr_name('citation')]
   end
 
+  def coauthor
```
lfarrell: Is coauthor used in any of the forms? I don't see it used in this PR

spec/features/create_work_spec.rb
smithjp: No, it look like I missed these when I was cleaning up.  I'll remove them.

app/models/solr_document.rb
smithjp: I'll check.  I think it was used for one of the work types created before we had the MAPs.  I'll remove it if it is no longer being used.

PR #11, opened 2018-02-01
Proquest import

smithjp:


lib/tasks/proquest_ingest.rake
```diff
@@ -0,0 +1,289 @@
...
+      end
+    end
+  end
+
+  def ingest_files(resource: nil, files: [], metadata: nil, zip_dir_files: nil)
+    ordered_members = []
+
+    files.each do |f|
+      file_path = zip_dir_files.find { |e| e.match(f) }
+      puts 'trying... '+f.to_s
```
bbpennel: This is mostly a style thing that I'm wondering about, just curious about the use of string concatenation instead of interpolation for logging in this script.  I'm not very experienced in this space, so I'd defer to you, I just had the impression that interpolation was more commonly used for this sort of stuff?  And might be more performant, but I don't think that would be significant in this case.  Also, does `.to_s` get called automatically in both cases?

lib/tasks/proquest_ingest.rake
```diff
@@ -0,0 +1,289 @@
...
+      unless embargo_release_date.blank?
+        visibility = visibility_during_embargo
+      end
+    end
+
+    title = metadata.xpath('//DISS_description/DISS_title').text
+
+    creators = metadata.xpath('//DISS_submission/DISS_authorship/DISS_author/DISS_name').map do |creator|
+      if creator.xpath('DISS_affiliation').text.eql? 'University of North Carolina at Chapel Hill'
+        creator.xpath('DISS_surname').text+', '+creator.xpath('DISS_fname').text+' '+creator.xpath('DISS_middle').text
```
bbpennel: Is this value trimmed?  Just imagine there will be a lot of records without a middle name, which will leave a trailing space.

lib/tasks/proquest_ingest.rake
```diff
@@ -0,0 +1,289 @@
...
+
+    keywords = metadata.xpath('//DISS_description/DISS_categorization/DISS_keyword').text.split(', ')
+    keywords << metadata.xpath('//DISS_description/DISS_categorization/DISS_category/DISS_cat_desc').text
+
+    abstract = metadata.xpath('//DISS_content/DISS_abstract').text
+
+    advisor = metadata.xpath('//DISS_description/DISS_advisor/DISS_name').map do |advisor|
+      advisor.xpath('DISS_surname').text+', '+advisor.xpath('DISS_fname').text+' '+advisor.xpath('DISS_middle').text
+    end
+    committee_members = metadata.xpath('//DISS_description/DISS_cmte_member/DISS_name').map do |advisor|
```
bbpennel: Would it be helpful to make a common method which formatted author names?

lib/tasks/proquest_ingest.rake
```diff
@@ -0,0 +1,289 @@
...
+    if language == 'en'
+      language = 'English'
+    end
+
+    file_full << metadata.xpath('//DISS_content/DISS_binary').text
+    file_full += metadata.xpath('//DISS_content/DISS_attachment').map do |file_name|
+      file_name.xpath('DISS_file_name').text
+    end
+
+    file.close
```
bbpennel: This file handle is just used to provide to nokogiri for parsing, might be clearer to close it right after parsing?

lib/tasks/proquest_ingest.rake
```diff
@@ -0,0 +1,289 @@
...
+
+    { resource: work_attributes, files: file_full }
+
+  end
+
+  def work_record(work_attributes)
+    resource = Dissertation.new
+    resource.creator = work_attributes['creator']
+    resource.depositor = @depositor_email
+    resource.save
```
bbpennel: Just curious about the reason for doing a `save` here rather than the end of the method?  And also why `save` instead of `save!` if not checking the value?  Is the hope that the `save!` after this method is called will succeed sometimes when this one fails?

lib/tasks/proquest_ingest.rake
```diff
@@ -0,0 +1,289 @@
...
+  require 'zip'
+
+  # set fedora access URL. replace with fedora username and password
+  # test environment will not have access to ERA's fedora
+
+  # Must include the email address of a valid user in order to ingest files
+  @depositor_email = 'admin@example.com'
+
+  # Use the ERA public interface to download original file and foxml
+  @fedora_url = ENV['FEDORA_PRODUCTION_URL']
```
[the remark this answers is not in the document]
smithjp: Everything is coming from zip files in a directory.  I think this must have come from the original migration task from Alberta.  I'll remove it.

lib/tasks/proquest_ingest.rake
```diff
@@ -0,0 +1,289 @@
...
+  FileUtils::mkdir_p @completed_dir
+
+
+  desc 'batch migrate generic files from FOXML file'
+  task :ingest, [:dir, :migrate_datastreams] => :environment do |t, args|
+    args.with_defaults(:migrate_datastreams => "true")
+
+    # Should deposit works into an admin set
+    # Update title parameter to reflect correct admin set
+    @admin_set_id = ::AdminSet.where(title: 'default').first.id
```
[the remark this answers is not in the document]
smithjp: Good idea.  I'll fix that.

lib/tasks/proquest_ingest.rake
smithjp: You are right that string interpolation is the preferred method for combining strings, I'm just in the habit of writing it this way for debugging.  I should also be using a rails logger statement for this if it is going to be used for non-development purposes.  I will update the ones here and in the other migration rake task.  As for the .to_s, I have found that it is usually called automatically, but I have run into problems in the past when I try to print or log non-string objects, so I usually add a .to_s or .as_json while I am debugging just to be safe.

lib/tasks/proquest_ingest.rake
smithjp: I wasn't thinking of that.  I'll update it to remove extra spaces.

lib/tasks/proquest_ingest.rake
smithjp: I think that this is another thing left over from the original rake task.  I will look at these again and see if the first one is necessary.

PR #12, opened 2018-02-01
Controlled vocabulary

smithjp:


lib/tasks/proquest_ingest.rake
```diff
@@ -0,0 +1,290 @@
...
+    resource.save
+
+    resource.label = work_attributes['title'][0]
+    resource.title = work_attributes['title']
+    resource.keyword =  work_attributes['keyword']
+    resource.degree_granting_institution = work_attributes['degree_granting_institution']
+    resource.abstract = [work_attributes['abstract']]
+    resource.advisor = work_attributes['advisor']
+    resource.degree = work_attributes['degree']
+    resource.academic_department = work_attributes['academic_department']
```
lfarrell: Hasn't this been replaced by affiliation?

lib/tasks/proquest_ingest.rake
```diff
@@ -0,0 +1,290 @@
...
+      genre = 'Dissertation'
+      resource_type = 'Dissertation'
+    else
+      genre = 'Thesis'
+      resource_type = 'Masters Thesis'
+    end
+
+    academic_concentration = metadata.xpath('//DISS_description/DISS_institution/DISS_inst_contact').text
+
+    affiliation = metadata.xpath('//DISS_description/DISS_institution/DISS_inst_contact').text.strip
```
lfarrell: Should this be in the work attributes?

lib/tasks/proquest_ingest.rake
smithjp: Yes.  It looks like I missed a few.

spec/fixtures/authorities/biomed_departments.yml
```diff
@@ -0,0 +1,10 @@
+## spec/fixtures/authorities/biomed_departments.yml
+terms:
+  - id: Biomedical Research Imaging Center, University of North Carolina at Chapel Hill
+    term: School of Medicine, Biomedical Research Imaging Center
+  - id: Biostatistics, Lineberger Comprehensive Cancer Center, University of North Carolina, Chapel Hill, NC, 27599-7295, USA
```
bbpennel: In the test it looks like the full address as expressed in one of the submission documents is mapped as a term.  Don't see that mapping in the non-fixture versions though.  How will this vocabulary handle addresses in the production version?  The addresses are also generally not all that consistently formatted unfortunately.

smithjp: These examples were copied from the biomed_departments.yml authority file in the config/authorities directory.  I was not sure what the addresses from biomed would really look like, so I just copied the mappings from the google doc.  Do I need to include an address formatting service as well?

PR #15, opened 2018-02-16
Bxc 2124 multimedia

harringj:
Hey, @smithjp! I thought I’d go ahead and create PR for this ticket so it will be easier for you to give me feedback. There are currently three tests failing, for reasons I don’t fully understand. And I’m not entirely sure I managed to get all of the needed metadata fields added back where they need to go. Anyway, any feedback would be much appreciated :-) Thanks!


smithjp:
It looks like you set everything up correctly.  The only issue I have found is that some of the OAI tests are failing.  Every time a new work type is added, the solr query in the beginning of the OAI endpoint tests (requests/oai_pmh_endpoint_spec.rb) needs to be updated with the name of the new work type.

It looks good, though!

PR #20, opened 2018-03-02
Scholarly works worktype

smithjp:


config/locales/scholarly_work.en.yml
```diff
@@ -0,0 +1,8 @@
+en:
+  hyrax:
+    icons:
+      scholarly_work:     'fa fa-file-text-o'
+    select_type:
+      scholarly_work:
+        name:               "Scholarly Work"
+        description:        "Scholarly work works"
```
bbpennel: As much as I enjoy "Scholarly work works", this and all the translations should probably have the extra "work" trimmed from them.

spec/controllers/hyrax/scholarly_works_controller_spec.rb
```diff
@@ -0,0 +1,9 @@
+# Generated via
+#  `rails generate hyrax:work ScholarlyWork`
+require 'rails_helper'
+
+RSpec.describe Hyrax::ScholarlyWorksController do
```
bbpennel: Looks like this can be deleted

spec/models/scholarly_work_spec.rb
```diff
@@ -0,0 +1,9 @@
+# Generated via
+#  `rails generate hyrax:work ScholarlyWork`
+require 'rails_helper'
+
+RSpec.describe ScholarlyWork do
+  it "has tests" do
```
bbpennel: Should this be filled out similar to the spec for https://github.com/UNC-Libraries/hy-c/blob/master/spec/models/honors_thesis_spec.rb ?

config/locales/scholarly_work.en.yml
smithjp: I agree.  I will ask Julie and Rebekah what the visible names and descriptions should be.

PR #23, opened 2018-03-27
Convert dates to edtf

smithjp:


smithjp:
This branch needs to be updated after the scholarly works branch is merged.

PR #26, opened 2018-04-05
Override admin set policies

smithjp:
I added a conditional checking if the current user is an admin in the interpret_visibility_actor file to allow the embargo release date restrictions from admin sets to be ignored.  I also found an error in the logic in the form_relationships file which reset the admin set to the default for the given work type even if an administrator had selected a different admin set.  I fixed that error here because I needed it working in order to test the override abilities.

### DIFF
--- a/app/actors/hyrax/actors/interpret_visibility_actor.rb
+++ b/app/actors/hyrax/actors/interpret_visibility_actor.rb
@@ -1,0 +1,238 @@
+# Overriding actor to allow admins to override admin set embargo permissions
+module Hyrax
+  module Actors
+    class InterpretVisibilityActor < AbstractActor
+      class Intention
+        def initialize(attributes)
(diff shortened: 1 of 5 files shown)


app/actors/hyrax/actors/interpret_visibility_actor.rb
```diff
@@ -0,0 +1,238 @@
...
+        return true if date.present? && date.future?
+
+        env.curation_concern.errors.add(attribute_name, "Must be a future date.")
+        false
+      end
+
+      # Validate an embargo date against permission template restrictions
+      def valid_template_embargo_date?(env, date, template)
+        # Added this to allow admins to override embargo settings in admin set
+        return true if env.current_ability.admin?
```
smithjp: This is the only real change to the file.  I also added comments to the validate method above (106).

PR #28, opened 2018-04-10
Homepage Redesign

lfarrell:


app/views/hyrax/homepage/_unc_deposit.html.erb
```diff
@@ -0,0 +1,27 @@
+<div id="unc-deposit" class="row text-center">
+  <h2><%= t('hyrax.deposit.header') %></h2>
+  <div class="col-sm-3 col-xs-6 unc-block">
+    <%= link_to hyrax.my_works_path do %>
```
smithjp: Should we go ahead and change this link to the create article link?  It may not be necessary at this point if we expect lots of changes to the homepage design.  Same for below.

app/views/hyrax/homepage/_unc_masthead.html.erb
```diff
@@ -0,0 +1,16 @@
+<div id="unc-search" class="row">
+  <div class="col-sm-offset-0 col-sm-6 col-xs-offset-1 col-xs-11">
+    <%= image_tag("UNC_logo_white.png", alt: t('hyrax.homepage_images.unc_logo')) %>
```
smithjp: This image is not showing up for me.  Has it been added to the repository?

lfarrell: Hmm, don't know what happened to it. I added it back in

smithjp:
This looks really good.

PR #29, opened 2018-04-26
Add rake task tests

smithjp:
Need to add libmediainfo and libzen libraries for fits to run in the tests.  I will also need to adjust sample data import task in the VM ansible scripts after this branch is merged.


smithjp:
This branch is ready for testing.  Make sure to update the VM with the code from update-fits-for-tests branch first.  You can run these tests alone with `scl enable rh-ruby24 -- bundle exec rspec ./spec/tasks` after running the test set up commands.

lib/tasks/proquest_ingest.rake
```diff
@@ -229,7 +228,7 @@ namespace :proquest do
         'creator'=>creators,
         'degree_granting_institution'=> degree_granting_institution,
         'keyword'=>keywords,
-        'abstract'=>abstract,
+        'abstract'=>abstract.gsub(/\n/, "").strip,
```
bbpennel: Were newlines in abstracts causing issues or were they not wanted for display purposes?

smithjp: I think this was just to clean it up for display.

PR #31, opened 2018-05-01
Shibboleth

lfarrell:
First pass at shibboleth integration, following DCE setup https://github.com/curationexperts/playbook/blob/master/authentication/shibboleth_and_hyrax.md


app/models/auth_config.rb
```diff
@@ -0,0 +1,8 @@
+class AuthConfig
+  # In production, use Shibboleth for user authentication,
+  # but in development mode, use local database
+  # authentication instead.
+  def self.use_database_auth?
+    !Rails.env.production? && ENV['DATABASE_AUTH'] == 'true'
```
smithjp: We may want to change this to only check the environment variable, otherwise we will not be able to run the application in production mode on the VM.  We can also try to run the application in development mode on the VM if that would be a better solution.

PR #33, opened 2018-05-25
Add travis ci

smithjp:
Need to update postgres ansible script for vagrant after this is merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ rvm:
   - 2.4
 before_script:
   - bundle exec rake db:create
-  - bundle exec rake hydra:test_server
+  - bundle exec rake hydra:test_server &
   - bundle exec rake test_setup RAILS_ENV=test
 before_install:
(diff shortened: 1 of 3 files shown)


bbpennel:
This looks good.

There are quite a few differences with the one used in hyrax (https://github.com/samvera/hyrax/blob/master/.travis.yml), which you are probably already aware of. Sorry to retread, just wanted to make sure I understand the differences:

* not using chrome in ours
* no postgres in theirs (I guess they are using a different database by default?)
* rvm version (2.5 vs 2.4)
* they appear to reference oraclejdk8 in `before_script` instead of via `jdk`, I'm not clear on why they would need to switch jdks in the same job
* no clamav in theirs, do they not run any clamav tests?
* they don't appear to be starting `rake hydra:test_server` in their .travis.yml, I wonder how they are instantiating the test environment? (the 75 second wait makes me slightly nervous :) )

PR #36, opened 2018-06-06
Map updates

smithjp:


app/views/records/edit_fields/_license.html.erb
```diff
@@ -0,0 +1,6 @@
+<% license_service = Hyrax::LicenseService.new %>
```
bbpennel: It looks like this is an override of:
https://github.com/samvera/hyrax/blob/master/app/views/records/edit_fields/_license.html.erb
So we should include an `[override]` note at the top

lib/tasks/proquest_ingest.rake
```diff
@@ -261,7 +261,7 @@ namespace :proquest do
     resource.abstract = [work_attributes['abstract']]
     resource.advisor = work_attributes['advisor']
     resource.degree = work_attributes['degree']
-    resource.affiliation = work_attributes['affiliation']
+    # resource.affiliation = work_attributes['affiliation']
```
bbpennel: Is this commented out, pending controlled vocab work?

PR #41, opened 2018-06-28
Upgrade to 2.1.0

smithjp:


db/migrate/20180618173454_create_collection_types.hyrax.rb
```diff
@@ -0,0 +1,17 @@
+class CreateCollectionTypes < ActiveRecord::Migration[5.0]
+  def change
+    create_table :hyrax_collection_types do |t|
```
bbpennel: More of a question, but are these migration scripts something that is useful to have checked in? Are they modified from the templates in the hyrax code base?

solr/config/solrconfig.xml
```diff
@@ -284,6 +284,8 @@
       -->
   </searchComponent>
 
+  <!-- TURN OFF SUGGEST - it causes performance issues -->
```
bbpennel: 👍 
One thing I was wondering, what is the difference between `solr/conf/` and `/solr/config`? The schemas do seem to differ, and one of them mentions being for a hydra demo and the other for blacklight. `solr/config` appears to be used by `solr_wrapper_test.yml` while i can't find any references in the hyc project to `solr/conf`. Just wondering if we want to keep both. If removing it would cause us to diverge in potentially negative ways i'm for keeping it, but it might be nice not to need to keep both versions up to date.

spec/features/create_dissertation_spec.rb
```diff
@@ -42,12 +42,12 @@
       DefaultAdminSet.create(work_type_name: 'Dissertation', admin_set_id: admin_set.id)
     end
 
-    scenario 'as a non-admin' do
-      login_as user
-
-      visit new_hyrax_dissertation_path
-      expect(page).to have_content "You are not authorized to access this page"
-    end
+    # scenario 'as a non-admin' do
```
bbpennel: Is this test no longer relevant?

solr/config/solrconfig.xml
smithjp: I think only the solr/config directory is being used, and the solr/conf directory can be removed.  The two directories were added to the repository by the gem's generators.

db/migrate/20180618173454_create_collection_types.hyrax.rb
smithjp: The migration files are almost all exact duplicates of the migration files in the gem.  They are required to be in our repository so that the database can be set up correctly in each environment.

solr/config/solrconfig.xml
bbpennel: well that certainly got your delete count up for this PR 👍

PR #43, opened 2018-07-05
Add DOI to citations

lfarrell:
Made "creator" a required field in general form, so citations show correctly for them.


app/helpers/hyrax/citations_behaviors/formatters/apa_formatter.rb
```diff
@@ -0,0 +1,86 @@
+module Hyrax
```
smithjp: Please add a comment like `<%# [hyc-override] Overriding partial in hyrax gem in order to add a link to admin dashboard %>` to the top of overridden gem files so we can keep track of our customizations and changes for future hyrax gem updates.

app/presenters/hyrax/article_presenter.rb
```diff
@@ -3,7 +3,7 @@
 module Hyrax
   class ArticlePresenter < Hyrax::WorkShowPresenter
     delegate :abstract, :access, :bibliographic_citation, :copyright_date, :date_captured,
-             :date_issued, :date_other, :doi, :edition, :extent, :funder, :genre, :geographic_subject, :issn,
```
smithjp: Does this need to be removed from the data set, dissertation, and general presenters as well?

app/presenters/hyrax/work_show_presenter.rb
```diff
@@ -0,0 +1,270 @@
+module Hyrax
```
smithjp: What was changed in this file?

smithjp:
The DOI attribute needs to be added to the forms and attribute_row files for each work type.  I am not able to add or view DOIs for some work types.

PR #47, opened 2018-07-09
Restrict sidekiq access

lfarrell:


lib/constraint/is_unc_admin.rb
```diff
@@ -0,0 +1,15 @@
+module Constraint
+  class IsUncAdmin
```
bbpennel: I wonder if it would make sense to name this class "IsAdmin" since none of the logic here is particular to UNC admins, and change the module to "Hyc" or something like that so that we can tell it is our code? That seems to be the approach in the only example I see in samvera https://github.com/samvera/hyrax/blob/1e504c200fd9c39120f514ac33cd42cd843de9fa/lib/generators/hyrax/templates/config/arkivo_constraint.rb

lib/constraint/is_unc_admin.rb
```diff
@@ -0,0 +1,15 @@
+module Constraint
```
bbpennel: Sorry i'm not as familiar with rails practices, was wondering if there was a reason for having this file in `/lib/constraint` all by itself rather than in the `/app` directory? The only stuff I see over there are rake tasks. The few examples I glanced at of custom constraint classes

smithjp: I think in general, any code not specific to the application can go in the /lib directory instead of the /app directory.  This usually includes generic data manipulations and potentially authorization.  I think it would be fine to leave this here.

smithjp:
I added the 404 code to this branch and saw the 404 page instead of a rails routes error when trying to access as a non-admin.

PR #57, opened 2018-07-18
Redirect or hide deposit links if user not logged in

lfarrell:


app/views/_controls.html.erb
```diff
@@ -6,12 +6,14 @@
           <%= link_to t(:'hyrax.links.collections'), hyrax.root_path %></li>
         <li <%= 'class=active' if current_page?(hyrax.root_path) %>>
           <%= link_to t(:'hyrax.links.departments'), hyrax.root_path %></li>
-        <li><%= link_to  '#', class: 'all-unc-work-types',
-                         data: { behavior: 'select-work', target: '#worktypes-to-create', 'create-type' => 'single' } do %>
-            <i class="fa fa-arrow-circle-up" aria-hidden="true"></i>
-            <%= t('hyrax.links.deposit') %>
-          <% end %>
-        </li>
+        <% if user_signed_in? %>
```
bbpennel: Why is the deposit link being omitted on the search result page when not logged in instead of forwarding to the login page like in all other cases??

lfarrell: We could, wasn't sure it was necessary to show, as it's only on internal pages.

bbpennel: My inclination would be to have the deposit buttons behave consistently (all of them show up before being logged in, all of them forward to the login page)

PR #59, opened 2018-07-18
differentiate between citations and downloads for citation services

lfarrell:
https://jira.lib.unc.edu/browse/HYC-168


smithjp:
If I click on the Zotero or Mendeley links and then go back, the citation modal is still up and cannot be closed.

PR #66, opened 2018-07-23
Add affiliations property back in

lfarrell:


lfarrell:
Combines https://jira.lib.unc.edu/browse/HYC-193 and https://jira.lib.unc.edu/browse/HYC-199

PR #68, opened 2018-07-24
Featured works on homepage

lfarrell:
The way it's setup at the moment we need to always have 3 featured works, so the spacing looks okay. It also centers everything vertically per Hong's suggestion.

### DIFF
--- a/app/views/hyrax/homepage/_featured.html.erb
+++ b/app/views/hyrax/homepage/_featured.html.erb
@@ -1,0 +1,7 @@
+<!-- [hyc-override] Overriding hyrax default -->
+<% presenter = featured.presenter %>
+
+<div class="col-xs-12 col-sm-6">
+  <%= render_thumbnail_tag presenter %>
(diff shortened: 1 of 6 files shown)


app/views/hyrax/homepage/_featured.html.erb
```diff
@@ -0,0 +1,7 @@
+<!-- [hyc-override] Overriding hyrax default -->
```
bbpennel: Switch to `<%# your_ruby_code %>` style comment so it doesn't show up in the browser

app/views/hyrax/homepage/_featured_works.html.erb
```diff
@@ -0,0 +1,4 @@
+<!-- [hyc-override] Overriding hyrax default -->
```
bbpennel: Switch to `<%# your_ruby_code %>` style comment so it doesn't show up in the browser

PR #72, opened 2018-07-27
Hyc-228 - Sign-in order and worktype visibility

bbpennel:
* Users forwarded to the page they requested or were on when they are prompted to log in to shibboleth.
* Modals for selecting worktype to deposit display options prior to logging in
* After selecting a worktype, the user will be prompted to log in and then sent to form
* Fixes issue where student papers modal displayed the full set of worktypes after visiting any page other than the front page.

This involved disabling turbolinks for login links, overriding hyrax default post-sign-in page, and passing `origin` parameter to shibboleth.


smithjp:
This looks great!

PR #73, opened 2018-07-31
Make sure datepicker works with fields that allow multiple dates and …

lfarrell:
…works without page reloading.


lfarrell:
https://jira.lib.unc.edu/browse/HYC-231

PR #75, opened 2018-07-31
Update properties for genre and resource type

lfarrell:
https://jira.lib.unc.edu/browse/HYC-232
https://jira.lib.unc.edu/browse/HYC-242
* Changes predicate for Resource Type field
* Removes genre field
* Adds dcmi_type field, which is not displayed on forms or record pages. It is provided a default value for all work types except General, Multimedia and Scholarly Works.
* Adds support for suppressing fields from forms such that no one can modify them (not even admins)
* Adds support for having fixed values for fields that override any user input.


app/models/article.rb
```diff
@@ -59,7 +59,7 @@ class Article < ActiveFedora::Base
     index.as :stored_searchable
   end
 
-  property :genre, predicate: ::RDF::Vocab::EDM.hasType do |index|
+  property :genre, predicate: ::RDF::URI('http://purl.org/dc/terms/type') do |index|
```
smithjp: This property is also in the RDF gem as DC.type (Vocab::DC.type).

smithjp:
The genre field is still labelled as genre in the form and on the show page.

app/forms/single_value_form.rb
```diff
@@ -22,7 +29,24 @@ def self.model_attributes(_)
         end
       end
     end
+    
+    # Insert fixed values, overriding any already defined values
+    fixed_term_values.each do |field, values|
```
smithjp: This is going to overwrite any migrated values when migrated works are edited.  Is this what we want to do?

app/forms/single_value_form.rb
```diff
@@ -22,7 +29,24 @@ def self.model_attributes(_)
...
+          attrs[field] = values.first
+        end
+      end
+    end
 
     attrs
   end
+  
+  def secondary_terms
+    super - suppressed_terms
```
smithjp: Should these fields be hidden from everyone or just from non-admins?

smithjp:
If we want the fedora-only fields to be hidden from the interface entirely, we will also want to remove them from the catalog controller.

bbpennel:
@smithjp After checking with Anna, it sounds like they should be visible to and modifiable by admins, and they should not override existing values. I probably misunderstood an earlier conversation. So this approach is not going to work. I will need to figure out how to modify the value in the model prior to populating the form rather than after. Providing the values at form render time was not working, although it may be possible if we wanted to override the behavior of multivalue input.

PR #80, opened 2018-08-01
Hyc 181 create users with permissions

smithjp:


app/views/accounts/new.html.erb
```diff
@@ -0,0 +1,23 @@
...
+  </div>
+</div>
+
+<div class="panel panel-default">
+  <div class="panel-body">
+    <div class="col-md-4">
+      <%= simple_form_for :account, url: accounts_create_path, method: :post do |f| %>
+        <div class="form-group">
+          <%= f.label :email %>
+          <%= f.text_field :email, pattern: '^\w{2,8}@\w*\.unc.edu$', title: 'UNC email address', class: 'form-control' %>
```
bbpennel: It looks like an onyen currently needs to be between 3-8 characters, but to be honest I would lean towards a `[a-zA-Z0-9]+` since the length policy may have changed at some point in the past or in the future, and also the `_` character is disallowed (unless you suspect that may not have always been the case). This should at least stop people from using `@unc.edu` aliases.

app/views/layouts/homepage.html.erb
```diff
@@ -1,3 +1,4 @@
+<%# [hyc-override] Overriding partial in hyrax gem to update homepage %>
```
bbpennel: I don't mind all these comments being on this PR, I just wanted to verify that they were intentional

spec/controllers/accounts_controller_spec.rb
```diff
@@ -0,0 +1,49 @@
+require 'rails_helper'
+
+RSpec.describe AccountsController, type: :controller do
```
bbpennel: It may be helpful to have a test that verifies that the controller can't be accessed by users without sufficient permissions.

app/views/layouts/homepage.html.erb
smithjp: Yes, I noticed that some were missing, so I went through all of the files and added the override comments.

PR #85, opened 2018-08-06
Hide user page from non-admins

lfarrell:
Just make sure user is authenticated and admin, instead of using constraints, https://jira.lib.unc.edu/browse/HYC-217


smithjp:
I can still access the users page as a non-admin

PR #87, opened 2018-08-08
Hy-C Facets

lfarrell:


smithjp:
The google doc had "type" (human_readable_type) listed as removed.  Is that still the case?  If we are not faceting rights statements, then everything else looks good.

lfarrell:
Updated. I was thinking human_readable type might break some of the other facets, but that doesn't seem to be the case.

PR #88, opened 2018-08-09
Don't display multiple affiliations on departments page

lfarrell:


bbpennel:
This changes the way that the affiliations are stored in Fedora, not just the way they are displayed on the full record page. If others are fine with it, then I guess that's okay, but is that what was being requested? I'm not sure what ticket this is part of.

PR #89, opened 2018-08-09
Allow admins to change visibility on admin sets

lfarrell:


bbpennel:
Just to make sure I am testing this correctly, this should make it so that when creating a work, an admin can select any of the visibility options without getting rejected for violating the template?

Not sure if you were getting this,  but as an admin in an admin set with "Public only" I can click on "Public" and "Embargo", but the other two options are disabled. As a deposit, all options are enabled, but I get rejected if I attempt to submit an invalid option.  I get this behavior on master as well, so I don't think its your change, but it does make it hard to test :)

bbpennel:
This does appear to work, but it's worth checking to see if we can do something about the state of the radio buttons. For admins, none should be disabled. For regular depositors, it should reflect the settings of the admin set. I don't know if this is an issue in Nurax or not, that's probably worth checking on or asking Rebekah to check on.

lfarrell:
Note: You have to add "Registered Users" as depositors in the admin set or this pull request won't work.

PR #96, opened 2018-08-14
Fix honors workflow

smithjp:
* Make affiliation required and single-valued in honors thesis form
* Remove requirement for academic concentration from honors thesis form
* Add feature test for reviewer group creation for honors thesis workflow


smithjp:
This is too messy

PR #97, opened 2018-08-15
Honors workflow

smithjp:
* Make affiliation required and single-valued in honors thesis form
* Remove requirement for academic concentration from honors thesis form
* Add feature test for reviewer group creation for honors thesis workflow

### DIFF
--- a/app/forms/single_value_form.rb
+++ b/app/forms/single_value_form.rb
@@ -7,10 +7,6 @@ class SingleValueForm < Hyrax::Forms::WorkForm
   # Map of fields to default values
   class_attribute :default_term_values
   self.default_term_values = Hash.new
-  class_attribute :admin_multi_value_fields
-  self.admin_multi_value_fields = Array.new
-  class_attribute :single_value_fields
-  self.single_value_fields = Array.new
 
   def initialize(model, current_ability, controller)
     initialize_default_term_values(model)
@@ -28,10 +24,6 @@ class SingleValueForm < Hyrax::Forms::WorkForm
 
   # cast single value fields back to multivalued so they will actually deposit
   def self.model_attributes(form_params)
-    admin_multi_value_fields.each do |field|
-      form_params[field] = Array(form_params[field])
-    end
-
     attrs = super
 
     single_value_fields.each do |field|
--- a/app/forms/single_value_form.rb
+++ b/app/forms/single_value_form.rb
@@ -7,7 +7,11 @@ class SingleValueForm < Hyrax::Forms::WorkForm
   # Map of fields to default values
   class_attribute :default_term_values
   self.default_term_values = Hash.new
-  
+  class_attribute :admin_multi_value_fields
+  self.admin_multi_value_fields = Array.new
+  class_attribute :single_value_fields
+  self.single_value_fields = Array.new
+
   def initialize(model, current_ability, controller)
     initialize_default_term_values(model)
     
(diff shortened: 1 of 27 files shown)


app/views/records/edit_fields/_affiliation.html.erb
```diff
@@ -1,17 +1,26 @@
...
+                input_html: { class: 'form-control' }
+    -%>
 <% else %>
-  <%= f.input :affiliation, as: :multi_value_select,
-              collection: DepartmentsService.select_all_options.sort,
-              include_blank: true,
-              item_helper: DepartmentsService.method(:include_current_value),
-              input_html: { class: 'form-control' }
-  -%>
+  <% if params[:affiliation] %>
```
bbpennel: Sorry to re-litigate the past, but I'm not clear on the rules/reasoning for rendering this field.
When creating a new work, I appear to get multiple single value selects for all the work types I checked, except Master's Papers which has a multivalued select. However, when you edit the masters paper, you get single value selects for all types. Is there a reason to not use single value selects in all cases (maybe related to the department selection page)? If so, perhaps we should add a comment?

app/views/records/edit_fields/_affiliation.html.erb
```diff
@@ -1,17 +1,26 @@
...
-<% if params[:affiliation] %>
-  <%= f.input :affiliation, as: :select,
-              collection: DepartmentsService.select_all_options.sort,
-              include_blank: true,
-              selected: params[:affiliation],
-              item_helper: DepartmentsService.method(:include_current_value),
-              input_html: { class: 'form-control', multiple: true }
-  -%>
+<% if (f.object.try(:admin_multi_value_fields)) && (f.object.admin_multi_value_fields.include? key) && !current_user.admin? %>
+    <%= f.input :affiliation, as: :select,
```
bbpennel: I think we may run into some issues with #98 here, since the affiliation is going to be broken into and stored as multiple values. So when the user edits their work, there is a high chance they will see the wrong affiliation (the parent department).
Also, if an admin adds multiple affiliations and then the depositor edits it, i think they will lose the other affiliations?

app/forms/single_value_form.rb
```diff
@@ -24,6 +28,10 @@ def self.multiple?(field)
 
   # cast single value fields back to multivalued so they will actually deposit
   def self.model_attributes(form_params)
+    admin_multi_value_fields.each do |field|
```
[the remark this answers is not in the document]
smithjp: The fields need to be converted to arrays if they were not multi-valued in the forms before the params are sanitized or they will not be permitted.  I will add comments about this in the code.  
These fields should show up as single-valued for non-admins when works are created.  I may need to adjust the form partials to allow the fields to be multi-valued for everyone on edit or not editable for non-admins on edit in order to retain all of the data.

smithjp:
It looks like #98 changed the affiliations form input box to pass the selected partial affiliation instead of the delimited full affiliation, so I adjusted the service to no longer try to split the strings.

spec/features/create_work_in_honors_thesis_workflow_spec.rb
```diff
@@ -0,0 +1,227 @@
...
+      AdminSet.create(title: ["honors thesis admin set"],
+                      description: ["some description"],
+                      edit_users: [user.user_key])
+    end
+
+    let(:permission_template) do
+      Hyrax::PermissionTemplate.create!(source_id: admin_set.id)
+    end
+
+    let(:honors_workflow) do
```
bbpennel: Is it not possible to load the honors workflow file rather than duplicating the policy here? Seems like it will be difficult to maintain like this

PR #98, opened 2018-08-17
Department Pages and facets should separate schools and departments

lfarrell:
https://jira.lib.unc.edu/browse/HYC-240


app/forms/single_value_form.rb
```diff
@@ -57,6 +57,19 @@ def self.model_attributes(form_params)
       end
     end
 
+    # Split out affiliations
+    unless !attrs.key?(:affiliation) || attrs[:affiliation].blank?
```
bbpennel: for readability, what would you think about making this `if attrs.key?(:affiliation) && !attrs[:affiliation].blank?` to avoid the double negative?

app/forms/single_value_form.rb
```diff
@@ -57,6 +57,19 @@ def self.model_attributes(form_params)
...
+      affiliations = []
+
+      attrs[:affiliation].each do |aff|
+        aff.split(';').each do |value|
+          affiliations.push(value)
```
bbpennel: Need to add some trimming of the value before adding it, there are leftover spaces after the ;'s which cause ` Coaching Education Minor` to be stored instead of `Coaching Education Minor`, so when you go back to edit the work the select boxes have invalid values.

lfarrell: Ah right, I thought of that over the weekend

smithjp:
The edit form is not populating the affiliation fields correctly.  When I try to edit a work with an affiliation that was split into 3 parts, I am only seeing one affiliation and two empty input fields.

PR #105, opened 2018-08-30
Add in text sanitizer, per https://github.com/samvera/hyrax/pull/3225

lfarrell:
Adding fix from hyrax pull request, since this is a file we override

### DIFF
--- a/app/helpers/hyrax/citations_behaviors/formatters/chicago_formatter.rb
+++ b/app/helpers/hyrax/citations_behaviors/formatters/chicago_formatter.rb
@@ -14,13 +14,13 @@ module Hyrax
           text = "<span class=\"citation-author\">#{text}</span>" if text.present?
           # Get Pub Date
           pub_date = setup_pub_date(work)
-          text << " #{pub_date}." unless pub_date.nil?
+          text << " #{whitewash(pub_date)}." unless pub_date.nil?
 
           text << format_title(work.to_s)
           pub_info = setup_pub_info(work, false)
-          text << " #{pub_info}." if pub_info.present?
+          text << " #{whitewash(pub_info)}." if pub_info.present?
           # UNC customization. Add DOI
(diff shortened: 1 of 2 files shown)


lfarrell:
Also see pull request https://github.com/samvera/hyrax/pull/3231

bbpennel:
This is merged into hyrax now at least, i'm not quite sure what to do about a bug fix from a version of hyrax we aren't on yet, but i guess it wouldn't do any harm

lfarrell:
Yes, but we won't pick up the changes since we've overridden the file. The joys of basically forking a file :)

app/helpers/hyrax/citations_behaviors/formatters/chicago_formatter.rb
```diff
@@ -14,13 +14,13 @@ def format(work)
...
+          text << " #{whitewash(pub_date)}." unless pub_date.nil?
 
           text << format_title(work.to_s)
           pub_info = setup_pub_info(work, false)
-          text << " #{pub_info}." if pub_info.present?
+          text << " #{whitewash(pub_info)}." if pub_info.present?
           # UNC customization. Add DOI
-          text << " #{work.doi[0]}" if !work.doi.nil? && work.doi.length > 0
+          text << " #{whitewash(work.doi[0])}" if !work.doi.nil? && work.doi.length > 0
           text.html_safe
```
smithjp: There are still instances of html_safe in each of the citation formatter files

lfarrell: They're that way in master too. I think the main issue was trusting user provided titles, at least as far as this request went.

lfarrell: This first three in this list should be fixed https://github.com/samvera/hyrax/issues/3187, by these changes and upgrading to 2.2.x
The fourth issue is still a problem https://github.com/samvera/hyrax/issues/3232

PR #107, opened 2018-08-30
Show file page thumbnails for public works that haven't been withdrawn

lfarrell:
State doesn't persist onto file pages, and only seems to exist in the context of a workflow. This should show/hide files correctly. The only exception is the file owner can still see thumbnails with a direct link to the files page, which is maybe okay.

### DIFF
--- a/app/views/hyrax/file_sets/media_display/_image.html.erb
+++ b/app/views/hyrax/file_sets/media_display/_image.html.erb
@@ -13,7 +13,8 @@
                   target: :_blank,
                   id: "file_download" %>
     </div>
-<% elsif @presenter.respond_to?('workflow') && !@presenter.workflow.in_workflow_state?(['withdrawn', 'pending deletion']) %>
+<% elsif (@presenter.respond_to?('workflow') && !@presenter.workflow.in_workflow_state?(['withdrawn', 'pending deletion'])) ||
+    !@presenter.respond_to?('workflow') %>
     <div>
       <%= image_tag thumbnail_url(file_set),
                     class: "representative-media",
@@ -22,11 +23,4 @@
     </div>
 <% elsif @presenter.respond_to?('workflow') && @presenter.workflow.in_workflow_state?(['withdrawn', 'pending deletion']) %>
   <%# Display nothing %>
-<% else %>
-  <div>
-    <%= image_tag thumbnail_url(file_set),
-                  class: "representative-media",
-                  alt: "",
-                  role: "presentation" %>
-  </div>
 <% end %>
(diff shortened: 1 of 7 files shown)


app/views/hyrax/file_sets/media_display/_image.html.erb
```diff
@@ -13,11 +13,20 @@
                   target: :_blank,
                   id: "file_download" %>
     </div>
-<% elsif !@presenter.workflow.in_workflow_state?(['withdrawn', 'pending deletion']) %>
+<% elsif @presenter.respond_to?('workflow') && !@presenter.workflow.in_workflow_state?(['withdrawn', 'pending deletion']) %>
```
bbpennel: I believe `!@presenter.respond_to?('workflow') || (@presenter.respond_to?('workflow') && !@presenter.workflow.in_workflow_state?(['withdrawn', 'pending deletion']))` would work here, and then the `else`  block could be removed

app/views/hyrax/file_sets/media_display/_office_document.html.erb
```diff
@@ -12,11 +13,20 @@
                   id: "file_download",
                   data: { label: file_set.id } %>
     </div>
-<% elsif !@presenter.workflow.in_workflow_state?(['withdrawn', 'pending deletion']) %>
+<% elsif presenter.respond_to?('workflow') && !@presenter.workflow.in_workflow_state?(['withdrawn', 'pending deletion']) %>
```
bbpennel: Can be probably be merged with the else block like in the other files

app/views/hyrax/file_sets/media_display/_pdf.html.erb
```diff
@@ -12,11 +13,20 @@
                   id: "file_download",
                   data: { label: file_set.id } %>
     </div>
-<% elsif !@presenter.workflow.in_workflow_state?(['withdrawn', 'pending deletion']) %>
+<% elsif presenter.respond_to?('workflow') && !@presenter.workflow.in_workflow_state?(['withdrawn', 'pending deletion']) %>
```
bbpennel: Possibly merge with the else block

app/views/hyrax/file_sets/media_display/_video.html.erb
```diff
@@ -13,12 +14,22 @@
                   target: :_blank,
                   id: "file_download" %>
     </div>
-<% elsif !@presenter.workflow.in_workflow_state?(['withdrawn', 'pending deletion']) %>
+<% elsif presenter.respond_to?('workflow') && !@presenter.workflow.in_workflow_state?(['withdrawn', 'pending deletion']) %>
```
bbpennel: Possibly merge with the else block

PR #115, opened 2018-09-05
Fix Google Analytics downloads parsing

lfarrell:


lfarrell:
Also requires https://gitlab.lib.unc.edu/cdr/vagrant-rails/merge_requests/28 to work correctly

lfarrell:
Yeah, looks like it's supposed to show views and downloads. I wonder how the downloads get wrapped in as they'll have a different url.

lfarrell:
Actually, I stand corrected. The downloads look they might only show up on the files page: https://ir-qa.lib.unc.edu/files/70795767w/stats?locale=en

PR #119, opened 2018-09-10
[HYC-144][HYC-297] Migration script

smithjp:
* example run command: `scl enable rh-ruby24 -- bundle exec rake cdr:migration:items['collection1','/spec/fixtures/migration_config.yml','/spec/fixtures/tasks/migration/mapping.csv']`
* also need to merge https://gitlab.lib.unc.edu/cdr/vagrant-rails/merge_requests/29 for rebuilding vm


lib/tasks/migrate.rake
```diff
@@ -104,7 +116,7 @@ namespace :cdr do
 
     def ingest_file(parent: nil, resource: nil, f: nil)
       file_set = FileSet.create(resource)
-      actor = Hyrax::Actors::FileSetActor.new(file_set, User.find_by_email(DEPOSITOR_EMAIL))
+      actor = Hyrax::Actors::FileSetActor.new(file_set, User.find_by_user_key(@depositor_key))
```
bbpennel: While I'm guessing this isn't expensive to look the depositor user up, would it be reasonable to hold onto the User object the first time it is looked up at line 19 and reuse it, since it doesn't seem like it will change per collection?

lib/tasks/migrate.rake
```diff
@@ -72,17 +82,19 @@ namespace :cdr do
...
-              csv << [uuid, resource.id]
-            end
+          @mapping[uuid] = resource.id
 
+          # Record old and new ids for works
+          CSV.open(@csv_output, 'a+') do |csv|
+            csv << [uuid, resource.id]
+          end
+
+          if !metadata_fields[:files].blank? && !metadata_fields[:files][0].blank?
```
bbpennel: I wouldn't mind a short comment here along the lines of "ingest files for this work if at least one is found"

lib/tasks/migrate.rake
```diff
@@ -151,130 +163,368 @@ namespace :cdr do
...
+        if !node.xpath('mods:namePart[@type="family"]', MigrationConstants::NS).text.blank?
+          arranger << node.xpath('concat(mods:namePart[@type="family"], ", ", mods:namePart[@type="given"])', MigrationConstants::NS)
+        else
+          arranger << node.xpath('mods:namePart', MigrationConstants::NS).text
+        end
+      end
+      degree_granting_institution = mods_version.xpath('mods:name[mods:role/mods:roleTerm/text()="Degree granting institution"]/mods:namePart', MigrationConstants::NS).map(&:text)
+      conference_name = mods_version.xpath('mods:name[@displayLabel="Conference" and @type="conference"]/mods:namePart', MigrationConstants::NS).map(&:text)
+      orcid = mods_version.xpath('mods:name/mods:nameIdentifier[@type="orcid"]', MigrationConstants::NS).map(&:text)
+      affiliation = mods_version.xpath('mods:name/mods:affiliation', MigrationConstants::NS).map(&:text)
```
bbpennel: Shouldn't there be an association between the named entity and their affiliation since we are going the user object route?

lib/tasks/migrate.rake
```diff
@@ -151,130 +163,368 @@ namespace :cdr do
         return
       end
 
-      title = mods_version.xpath('mods:titleInfo//mods:title', MigrationConstants::NS).text
+      title = mods_version.xpath('mods:titleInfo[not(@*)]/mods:title', MigrationConstants::NS).map(&:text)
```
bbpennel: How would you feel about splitting the entire MODS mapping into a separate file? Or at least method if that's not practical. I feel like it would be easier to read and also be easier to share with metadata people so that they could review the mappings. The values could go into a Hash that gets merged into work_attributes later.

lib/tasks/migrate.rake
```diff
@@ -151,130 +163,368 @@ namespace :cdr do
...
+        if !node.xpath('mods:namePart[@type="family"]', MigrationConstants::NS).text.blank?
+          advisor << node.xpath('concat(mods:namePart[@type="family"], ", ", mods:namePart[@type="given"])', MigrationConstants::NS)
+        else
+          advisor << node.xpath('mods:namePart', MigrationConstants::NS).text
+        end
+      end
+      funder_node = mods_version.xpath('mods:name[mods:role/mods:roleTerm/text()="Funder"]', MigrationConstants::NS)
+      funder = []
+      funder_node.each do |node|
+        if !node.xpath('mods:namePart[@type="family"]', MigrationConstants::NS).text.blank?
```
bbpennel: Constructing a persons name is pretty complex and applies to a lot of fields (as proven by this PR). I would suggest breaking the construction of the name out into a helper method. It looks like you may be able to even make a helper method that takes a role and adds the name to a provided array, which should make this section a lot shorter.

bbpennel: After talking some with Anna, it sounds like we also need to check for `type="personal"` and `type="corporate"`. If it is personal, then do the type="family" route. If it is corporate, then just expect a namePart without an attribute and use that as it is.
If a personal name doesn't have subsections, I think it might be worth throwing an exception and logging the issue (maybe cancelling ingest of the object for examination and ingest later).

lib/tasks/migrate.rake
```diff
@@ -151,130 +163,368 @@ namespace :cdr do
...
 
-      rdf_version = metadata.xpath("//foxml:xmlContent//rdf:RDF", MigrationConstants::NS).last
-      if rdf_version.to_s.match(/contains/)
-        contained_files = rdf_version.xpath("rdf:Description/*[local-name() = 'contains']/@rdf:resource", MigrationConstants::NS)
-        contained_files.each do |contained_file|
-          tmp_uuid = get_uuid_from_path(contained_file.to_s)
-          file_full << @object_hash[tmp_uuid]
+        if rdf_version.to_s.match(/hasModel/)
+          cdr_model_type = rdf_version.xpath('rdf:Description/*[local-name() = "hasModel"]/@rdf:resource', MigrationConstants::NS).map(&:text)
+          cdr_model_type = (cdr_model_type.include? 'info:fedora/cdr-model:AggregateWork') ? 'aggregate' : ''
```
bbpennel: what other types is this going to encounter? It seems like any types other than Aggregate would be lost.

lib/tasks/migrate.rake
```diff
@@ -151,130 +163,368 @@ namespace :cdr do
...
+      end_page = mods_version.xpath('mods:relatedItem[@type="host"]/mods:part/mods:extent[@unit="page"]/mods:end',MigrationConstants::NS).map(&:text)
+      related_url = mods_version.xpath('mods:relatedItem/mods:location/mods:url',MigrationConstants::NS).map(&:text)
+      url = mods_version.xpath('mods:location/mods:url',MigrationConstants::NS).map(&:text)
+      publisher_version = mods_version.xpath('mods:location/mods:url[@displayLabel="Publisher Version"] | mods:relatedItem[@type="otherVersion"]/mods:location',MigrationConstants::NS).map(&:text)
+      digital_collection = mods_version.xpath('mods:relatedItem[@displayLabel="Collection" and @type="host"]/mods:titleInfo/mods:title',MigrationConstants::NS).map(&:text)
+
+
+      deposit_record = ''
+      cdr_model_type = ''
+      rdf_version = metadata.xpath("//rdf:RDF", MigrationConstants::NS).last
```
bbpennel: Could you go directly to `//rdf:RDF/rdf:Description` here and then not need to include `rdf:Description/` in all the individual property xpath expressions?

lib/tasks/migrate.rake
```diff
@@ -151,130 +163,368 @@ namespace :cdr do
...
+            if !@binary_hash[tmp_uuid].blank?
+              file_full << @object_hash[tmp_uuid]
+            elsif !@object_hash[tmp_uuid].blank? && tmp_uuid != uuid
+              @parent_hash[uuid] << tmp_uuid
+            end
+          end
 
-      publisher = mods_version.xpath('mods:originInfo//mods:publisher',MigrationConstants::NS).text
-      language = mods_version.xpath('mods:language//mods:languageTerm',MigrationConstants::NS).text
+          if file_full.count > 1
```
bbpennel: Could you add a comment describing what is happening in this block?

lib/tasks/migrate.rake
```diff
@@ -151,130 +163,368 @@ namespace :cdr do
...
 
-      if rdf_version.to_s.match(/metadata-patron/)
-        patron = rdf_version.xpath("rdf:Description/*[local-name() = 'metadata-patron']", MigrationConstants::NS).text
-        if patron == 'public'
-          if rdf_version.to_s.match(/contains/)
-            visibility = Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PUBLIC
-          else
+        # Set default visibility first
+        visibility = Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PUBLIC
+        if rdf_version.to_s.match(/metadata-patron/)
```
bbpennel: A comment describing that the next block is about setting access controls may be helpful, or breaking it out into a helper method.

lib/tasks/migrate.rake
```diff
@@ -151,130 +163,368 @@ namespace :cdr do
...
       end
 
-      if language == 'eng'
-        language = 'English'
+      if !language.blank?
+        language.map!{|e| LanguagesService.label("http://id.loc.gov/vocabulary/iso639-2/#{e.downcase}") ?
+                              "http://id.loc.gov/vocabulary/iso639-2/#{e.downcase}" : e}
+      end
+
+      collection = Collection.where(title: @collection_name).first
```
bbpennel: Comment introducing this section or splitting into a helper method would be useful

lib/tasks/migrate.rake
```diff
@@ -151,130 +163,368 @@ namespace :cdr do
...
+          'contributor'=>contributor,
+          'creator'=>creator,
           'date_created'=>(Date.try(:edtf, date_created) || date_created).to_s,
-          'keyword'=>keywords,
           'date_modified'=>(Date.try(:edtf, date_modified) || date_modified).to_s,
-          'contributor'=>contributors,
-          'description'=>[description],
+          'description'=>description,
           'identifier'=>identifier,
+          'keyword'=>keywords,
```
bbpennel: soooo many

lib/tasks/migrate.rake
```diff
@@ -151,130 +163,368 @@ namespace :cdr do
...
 
-        if file_full.count > 1
-          representative = rdf_version.xpath('rdf:Description/*[local-name() = "defaultWebObject"]/@rdf:resource', MigrationConstants::NS).to_s.split('/')[1]
-          if representative
-            representative = @object_hash[get_uuid_from_path(representative)]
-            file_full -= [representative]
-            file_full = [representative] + file_full
-          end
+        if cdr_model_type == 'aggregate'
+          @parent_hash[uuid] = Array.new
```
bbpennel: Would changing the initialization of parent_hash to`@parent_hash = Hash.new {|h,k| h[k] = Array.new }` fulfill the objective of this line?

lib/tasks/migration_config.yml
```diff
@@ -5,9 +5,62 @@ collection1:
...
   collection_list: spec/fixtures/migration/collection.csv
   objects: spec/fixtures/migration/objects.txt
   binaries: spec/fixtures/migration/binaries.txt
   work_type: General
-  admin_set: default
\ No newline at end of file
+  admin_set: default
+  depositor_email: admin@example.com
+  collection_name: New Collection
+collection3:
```
bbpennel: hmm, i don't quite understand what this file is. Are these eventually going to be the actual migration tasks for individual collections/work_type/admin_set combinations? Will they get more descriptive names later?

lib/tasks/migrate.rake
```diff
@@ -1,64 +1,74 @@
...
-          key = get_uuid_from_path(value)
-          @binary_hash[key] = value
+    task :items, [:collection, :configuration_file, :mapping_file] => :environment do |t, args|
+      puts Time.now
+      if AdminSet.where(title: ENV['DEFAULT_ADMIN_SET']).count != 0
+        config = YAML.load_file(args[:configuration_file])
+        collection_config = config[args[:collection]]
+        @work_type = collection_config['work_type']
+        @admin_set = collection_config['admin_set']
+        @depositor_key = User.where(email: collection_config['depositor_email']).first.uid
```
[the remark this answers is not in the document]
smithjp: There are  scripts for adding an admin set and a user which are usually run when the application is initialized.  Right now, it is assumed that the necessary admin sets and preferred user have been created, but I can add some checks and documentation to make sure that everything is set up correctly.

lib/tasks/migration_config.yml
smithjp: This is just an example configuration file.  I am expecting to have more descriptive names for the collections when I update the configuration file for migrating real data.

lib/tasks/migrate.rake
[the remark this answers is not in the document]
bbpennel: I think documenting the process would be sufficient

smithjp:
The instructions for running the migration script are on confluence: https://confluence.lib.unc.edu/display/RAT/Migration+Workflow

lib/tasks/migrate.rake
```diff
@@ -10,18 +12,20 @@ namespace :cdr do
 
     desc 'batch migrate generic files from FOXML file'
     task :items, [:collection, :configuration_file, :mapping_file] => :environment do |t, args|
-      puts Time.now
+
+      puts '['+Time.now.to_s+'] Start migration of '+args[:collection]
```
bbpennel: While its mainly a style decision, it seems like this would be easier to do with string interpolation? `puts "#{Time.now.to_s} Start migration of #{args[:collection]}"`

lib/tasks/migrate/services/mods_parser.rb
```diff
@@ -0,0 +1,258 @@
+module Migrate
+  module Services
+    class ModsParser
```
bbpennel: You might want to rename this "MetadataParser" since its handling the entire foxml document

lib/tasks/migrate/services/mods_parser.rb
```diff
@@ -0,0 +1,258 @@
...
+        child_works = Array.new
+
+        # get the uuid of the object
+        uuid = get_uuid_from_path(metadata.at_xpath('foxml:digitalObject/@PID', MigrationConstants::NS).value)
+        puts "getting metadata for: #{uuid}"
+
+        work_attributes = Hash.new
+
+        representative = ''
+        work_attributes['visibility_during_embargo'] = Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PRIVATE
```
bbpennel: Just curious, do these need to always be set, or could they just be set when you know there is an embargo on the object?

smithjp: I think they can be set only when there is an embargo.

PR #121, opened 2018-09-11
Update Hyrax to 2.2.3

lfarrell:
The updates appear to be pretty minor. I only found one issue with links displaying incorrectly. It's not nearly as involved as the update from 2.0 to 2.1. I'll double check the commits to see if there's anything we've overridden.


smithjp:
2.2.4 was released on 9/11/18 with some important bugfixes, so I will update this branch to switch to the latest version.

PR #131, opened 2018-09-28
Fix Resource Type Not Being Auto Selected for Work Types

lfarrell:
https://jira.lib.unc.edu/browse/HYC-287


lfarrell:
Added in a "selected" check in the feature tests

PR #132, opened 2018-10-01
Get rid of jquery datepicker and use native HTML 5 date elements

lfarrell:


lfarrell:
https://jira.lib.unc.edu/browse/HYC-300

PR #133, opened 2018-10-02
Hide general, dissertation and art mfa worktypes from non-admins

lfarrell:
https://jira.lib.unc.edu/browse/HYC-303


lfarrell:
Ended up removing that check. Redone so it's not needed

PR #136, opened 2018-10-02
Switch to onyen instead of email for new users

lfarrell:
https://jira.lib.unc.edu/browse/HYC-301

### DIFF
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -7,7 +7,7 @@ class AccountsController < ApplicationController
   end
 
   def create
-    notice = create_user(account_params[:email]).strip
+    notice = create_user(account_params[:onyen]).strip
 
     respond_to do |format|
       format.html { redirect_to hyrax.admin_users_path, notice: notice }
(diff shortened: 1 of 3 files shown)


app/controllers/accounts_controller.rb
```diff
@@ -17,15 +17,15 @@ def create
...
   def ensure_admin!
     authorize! :read, :admin_dashboard
   end
 
-  def create_user(email)
-    onyen = email.split('@').first
+  def create_user(onyen)
+    email = "#{onyen}@ad.unc.edu"
     if User.where(uid: onyen).blank?
       user = User.where(provider: 'shibboleth', uid: onyen, email: email).first_or_create
```
smithjp: This query needs to be changed to `User.where(uid: onyen).first_or_create(provider: 'shibboleth', email: email)`
Also the query on line 32 of the user model needs to be changed similarly. 

I think those might be the only two queries that need to change.

bbpennel:
Have we tested this on QA yet?

lfarrell:
Not that I'm aware of

bbpennel:
I think we should try it there when it is available for a deploy, since we can't really test this locally.

smithjp:
We just tested this on QA, and it seems to be working.

PR #139, opened 2018-10-04
Update default admin set group/role management text

lfarrell:
Update default admin set group/role management text to match the dashboard links. Had to pull in the roles partials from hydra-role-management gem for some reason to get the overrides to take.
https://jira.lib.unc.edu/browse/HYC-304


bbpennel:
Could you add override comments to each of the files pulled in, even if you didn't make any changes to them?

PR #140, opened 2018-10-04
Make sure "Withdraw" button displays for admins on deposited works

lfarrell:
https://jira.lib.unc.edu/browse/HYC-253


app/views/hyrax/base/_show_actions.html.erb
```diff
@@ -17,9 +17,9 @@
     <% if current_user.admin? %>
       <%= link_to "Delete", [main_app, presenter], class: 'btn btn-danger', data: { confirm: "Delete this #{presenter.human_readable_type}?" }, method: :delete %>
       <% if @presenter.workflow.actions.present? &&
-          !@presenter.workflow.actions.select{|action| action[0] == 'withdraw'}.blank? %>
+          !@presenter.workflow.actions.select{|action| action[0] == 'withdrawn'}.blank? %>
```
bbpennel: Instead of changing these, the workflows should be updated to name the action 'Withdraw'. I think that was the intention of one of the previous PRs, but we missed updating the name there. This shouldn't require migrating any works, just running the workflow update task.

PR #143, opened 2018-10-05
Check for correct art work label

lfarrell:
Small fix to hide Art Work form. It was checking for the wrong string value

### DIFF
--- a/app/views/shared/_select_work_type_modal.html.erb
+++ b/app/views/shared/_select_work_type_modal.html.erb
@@ -8,7 +8,7 @@
       </div>
       <div class="modal-body">
         <% create_unrestricted_work_presenter.each do |row_presenter| %>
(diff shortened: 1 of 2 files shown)


app/views/shared/_select_work_type_modal.html.erb
```diff
@@ -8,7 +8,7 @@
       </div>
       <div class="modal-body">
         <% create_unrestricted_work_presenter.each do |row_presenter| %>
-          <% if (row_presenter.concern.to_s != 'Dissertation' && row_presenter.concern.to_s != 'General' && row_presenter.concern.to_s != 'ArtMfa') ||
```
bbpennel: Since we never want to show the artwork type, how about just suppressing it here and removing that js bit?

`if row_presenter.concern.to_s != 'ArtWork' && (current_user&.admin? || (row_presenter.concern.to_s != 'Dissertation' && row_presenter.concern.to_s != 'General')) %>

PR #144, opened 2018-10-08
Accessibility updates

lfarrell:
https://jira.lib.unc.edu/browse/HYC-307
Not sure what we can do about this one:

Change #addfiles to have a more visible :focus state
For some reason focusing it with tab won't change its appearance, I think it has to do with <input> being wrapped within <span>, so the it's tabbing through <input> instead of the <span>
Adding tabindex="1" attribute on the <span> itself fixes this, but it won't respond to user action because the file upload function still tied to the <input> within it, so my guess some script needs to be used to modify this behavior.

Technically it is accessible with the keyboard. If focus is set as above you have to tab twice and hit enter, which is highly counter-intuitive. If you don't set a tabindex it doesn't focus, but the file folder selector does pop open when enter is hit after one tab.


lfarrell:
Also left these out as page is really just a regular Blacklight search

Add a <h1> tag of "Browse Collections | Carolina Digital Repository"
I am not sure if this can work because it's a URL with specific parameters and we might not be able to just set the <h1> for this view; if not then use "Browse Contents"
Change <title> to "Browse Collections | Carolina Digital Repository"
Also related to the point above: not sure if we can just set for this one

scorchingoyster:
> Technically it is accessible with the keyboard. If focus is set as above you have to tab twice and hit enter, which is highly counter-intuitive. If you don't set a tabindex it doesn't focus, but the file folder selector does pop open when enter is hit after one tab.

Yeah, it is still accessible by keyboard. I was just checking it against the keyboard accessible guidelines and the visual state change isn't obvious enough. If we can't come to a solution then let's set this aside for now; the more important thing is that the buttons are sufficiently usable by screen readers.

Speaking of which, I am testing testing this again using NVDA tabbing through the upload buttons readers "Browse... button", without reading out text in the `<span>` tag (`<span>Add files...</span>`), which means it could be challenging for screen reader users to figure out what the buttons do.

I am not sure what's causing this so I need some time to look into it.

PR #148, opened 2018-10-08
Reassign reviewer after depositor finishes making requested changes.

lfarrell:
https://jira.lib.unc.edu/browse/HYC-285
I think this should work. Would need to test on QA


bbpennel:
This looks fine to me on reading, but we should test it in QA, like you said.

smithjp:
What ticket is this?

PR #149, opened 2018-10-08
Text updates for art work type

lfarrell:
https://jira.lib.unc.edu/browse/HYC-311
https://jira.lib.unc.edu/browse/HYC-312


bbpennel:
Clarifying the use case on the ticket before reviewing

PR #151, opened 2018-10-09
Integrate Browse Everything

lfarrell:
https://jira.lib.unc.edu/browse/HYC-38

Filesystem and DropBox are setup

Depends on https://gitlab.lib.unc.edu/cdr/vagrant-rails/merge_requests/33


app/views/hyrax/base/_form_files.html.erb
```diff
@@ -0,0 +1,63 @@
+<%# [hyc-override] Overriding partial to get browse everything to work in merged metadata/forms tabe %>
```
bbpennel: typo, "tab"

lfarrell: Hmm, should have been fixed here https://github.com/UNC-Libraries/hy-c/blob/13bc64e6fa62610f0e44218137720e849a971569/config/browse_everything_providers.yml#L9. Wonder if you need to restart?

lfarrell: You'll also need to be running this branch of the VM https://gitlab.lib.unc.edu/cdr/vagrant-rails/merge_requests/33

lfarrell: If that doesn't work you can manually add them in local_env.yml

PR #152, opened 2018-10-09
Setup geonames

lfarrell:
https://jira.lib.unc.edu/browse/HYC-241
Depends on: https://gitlab.lib.unc.edu/cdr/vagrant-rails/merge_requests/34


lfarrell:
It should work on the vm too.

PR #153, opened 2018-10-09
Restrict which fields to show in search results

lfarrell:
Restrict which fields to show in search results and truncate long field values
https://jira.lib.unc.edu/browse/HYC-313

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -72,25 +72,10 @@ class CatalogController < ApplicationController
     # solr fields to be displayed in the index (search results) view
     #   The ordering of the field names is the order of the display
     config.add_index_field solr_name("title", :stored_searchable), label: "Title", itemprop: 'name', if: false
-    config.add_index_field solr_name("description", :stored_searchable), itemprop: 'description', helper_method: :iconify_auto_link
-    config.add_index_field solr_name("keyword", :stored_searchable), itemprop: 'keywords', link_to_search: solr_name("keyword", :facetable)
-    config.add_index_field solr_name("subject", :stored_searchable), itemprop: 'about', link_to_search: solr_name("subject", :facetable)
     config.add_index_field solr_name("creator", :stored_searchable), itemprop: 'creator', link_to_search: solr_name("creator", :facetable)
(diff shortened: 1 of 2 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -72,25 +72,10 @@ def self.modified_field
...
-    config.add_index_field solr_name("proxy_depositor", :symbol), label: "Depositor", helper_method: :link_to_profile
-    config.add_index_field solr_name("depositor"), label: "Owner", helper_method: :link_to_profile
-    config.add_index_field solr_name("publisher", :stored_searchable), itemprop: 'publisher', link_to_search: solr_name("publisher", :facetable)
-    config.add_index_field solr_name("based_near_label", :stored_searchable), itemprop: 'contentLocation', link_to_search: solr_name("based_near_label", :facetable)
-    config.add_index_field solr_name("language", :stored_searchable), helper_method: :language_links, link_to_search: solr_name("language", :facetable)
-    config.add_index_field solr_name("date_uploaded", :stored_sortable, type: :date), itemprop: 'datePublished', helper_method: :human_readable_date
-    config.add_index_field solr_name("date_modified", :stored_sortable, type: :date), itemprop: 'dateModified', helper_method: :human_readable_date
     config.add_index_field solr_name("date_created", :stored_searchable), itemprop: 'dateCreated'
```
smithjp: Do we want date_created or date_issued here?

smithjp:
Something strange is happening with truncated creator names.  I think the creator name is being linked to a facet search which does not work well with truncating.  The link name will need to be truncated instead of the whole html element.
screen shot 2018-10-11 at 12 45 37 pm

app/controllers/catalog_controller.rb
lfarrell: Seems like they're not used consistently, but it was what was in the ticket. I'll check with Rebecca.

lfarrell: Fun, hand't thought about it truncating the URL

lfarrell: Should be fixed now. Doesn't make much sense to make it a link if you can't search all the authors.

lfarrell:
Bad force push that missed some changes. Just easier to open a new request

PR #161, opened 2018-10-18
Rename ArtWork work type to Artwork

lfarrell:
https://jira.lib.unc.edu/browse/HYC-320


bbpennel:
I think you will also need to update the various artwork.<lang>.yml files to change 
```
de:
  hyrax:
    icons:
      art_work:     'fa fa-file-text-o'
```
to `artwork:     'fa fa-file-text-o'`

PR #162, opened 2018-10-18
Hyc-258 Longleaf registration

bbpennel:
https://jira.lib.unc.edu/browse/HYC-258

Causes files to be registered in longleaf during ingest


bbpennel:
@smithjp no, that was very much an accident caused by a bad copy and paste (accidentally put a .git directory into tmp, so it decided it was a submodule). It should be gone now.

bbpennel:
@smithjp Sorry about that, lost some bits in my messy working directory from attempting to figure out how to setup QA. The VM should build and work properly now

PR #177, opened 2018-11-01
enable FAST integration

lfarrell:
https://jira.lib.unc.edu/browse/HYC-346


lfarrell:
Yeah, I don't think we want the required parameter. Removed it.

PR #178, opened 2018-11-01
Multimedia work form updates

lfarrell:
https://jira.lib.unc.edu/browse/HYC-342


lfarrell:
Displays all three form tabs at once for unknown reasons. Just making a new request instead

PR #187, opened 2018-11-09
Make rights statement single valued everywhere

lfarrell:
https://jira.lib.unc.edu/browse/HYC-347

### DIFF
--- a/app/actors/hyrax/actors/base_actor.rb
+++ b/app/actors/hyrax/actors/base_actor.rb
@@ -1,0 +1,105 @@
+# [hyc-override] Overriding actor to cast rights statements as single valued
+module Hyrax
+  module Actors
+    ##
+    # Defines the basic save/destroy and callback behavior, intended to run
+    # near the bottom of the actor stack.
+    #
+    # @example Defining a base actor for a custom work type
+    #   module Hyrax
+    #     module Actors
+    #       class MyWorkActor < Hyrax::Actors::BaseActor; end
+    #     end
+    #   end
+    #
+    # @see Hyrax::Actor::AbstractActor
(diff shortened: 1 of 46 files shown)


app/forms/hyrax/article_form.rb
```diff
@@ -16,7 +16,7 @@ class ArticleForm < ::SingleValueForm
 
     self.terms -= [:contributor, :based_near, :source, :description]
     
-    self.single_value_fields = [:title, :license, :rights_statement]
+    self.single_value_fields = [:title, :license]
```
bbpennel: Is rights_statement being removed from these because it already presents as a single valued field by default in hyrax? Its kind of too bad we can't just make the field un-special and handle it like any other single value field

lfarrell: Yeah, it should be removed from every form, as it's now single valued by default.

app/actors/hyrax/actors/base_actor.rb
```diff
@@ -0,0 +1,105 @@
...
+        env.curation_concern.date_modified = TimeService.time_in_utc
+      end
+
+      # Cast any singular values from the form to multiple values for persistence
+      # also remove any blank assertions
+      # TODO this method could move to the work form.
+      def clean_attributes(attributes)
+        attributes[:license] = Array(attributes[:license]) if attributes.key? :license
+        # [hyc-override] Overriding actor to cast rights statements as single valued
+        attributes[:rights_statement] = attributes[:rights_statement] if attributes.key? :rights_statement
```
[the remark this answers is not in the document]
lfarrell: Updated

PR #195, opened 2018-11-20
People objects

smithjp:


app/actors/hyrax/actors/general_actor.rb
```diff
@@ -3,6 +3,18 @@
 module Hyrax
   module Actors
     class GeneralActor < Hyrax::Actors::BaseActor
+
+      protected
+
+      def save(env)
+        env.attributes.each do |k,v|
```
bbpennel: Could you add a comment explaining what the purpose of this save action is?

bbpennel:
There aren't any `*` indicating which fields in a person block are required. Also, fields that were required before are not being required anymore, like `creator`

app/assets/javascripts/person_objects.js
```diff
@@ -0,0 +1,120 @@
...
+});
+
+
+function attach_add_person_listeners(selector){
+    var add_selector = '#add-another-'+selector;
+    var remove_selector = '.remove-'+selector;
+    var cloning_row = '#'+selector+'-cloning_row';
+
+    $(cloning_row).find('input').each(function() {
+        if ($(this).prop('id').match('_id')) {
```
lfarrell: I'd only select this once. `var self = $(this)` and then `self.prop('whatever')...

app/assets/javascripts/person_objects.js
```diff
@@ -0,0 +1,120 @@
...
+    var remove_selector = '.remove-'+selector;
+    var cloning_row = '#'+selector+'-cloning_row';
+
+    $(remove_selector).on('click', function (event) {
+        // get current_index again and store it in local variable, decrement it and store it
+        var current_index = $('#index-'+selector).val();
+        current_index--;
+
+        if (current_index == 1) {
+            $('.remove-'+selector).addClass('hidden');
```
lfarrell: same here with only selecting it once `var remove = $('.remove-' + selector)`

app/assets/javascripts/person_objects.js
```diff
@@ -0,0 +1,120 @@
...
+        } else {
+            $('.remove-'+selector).removeClass('hidden');
+        }
+
+        var name_input = $(cloning_row+' > .row').find('div.'+selector+'-name input').prop('name');
+        var model = name_input.split('[')[0];
+        var index = $(this).data('index');
```
lfarrell: Select this only once here, too.

app/assets/javascripts/person_objects.js
```diff
@@ -0,0 +1,120 @@
...
+        // stop page from reloading
+        event.preventDefault();
+
+        $(remove_selector).removeClass('hidden');
+
+        var current_index = $('#index-'+selector).val();
+
+        var $new_row = $(cloning_row+' > .row').clone();
+
+        var $name_input = $new_row.find('div.'+selector+'-name input');
```
lfarrell: Might want to break up all of these similar calls. You could make one large function and pass each values in `addRow('name')`, etc. or you could split each subsection out into an object method and return "this" to do method chaining `val.new_input().new_name().old_name()` etc. 

Let me know if that's not clear.

app/models/masters_paper.rb
```diff
@@ -22,11 +22,11 @@ class MastersPaper < ActiveFedora::Base
     index.as :stored_searchable
   end
 
-  property :advisor, predicate: ::RDF::URI('http://id.loc.gov/vocabulary/relators/ths') do |index|
-    index.as :stored_searchable, :facetable
```
lfarrell: No more faceting on advisor?

app/forms/hyrax/article_form.rb
```diff
@@ -6,11 +6,10 @@ class ArticleForm < ::SingleValueForm
...
 
     self.model_class = ::Article
-    self.terms += [:resource_type, :abstract, :access, :affiliation, :affiliation_label, :bibliographic_citation,
-                   :copyright_date, :date_issued, :date_other, :dcmi_type, :doi, :edition, :extent,
-                   :funder, :geographic_subject, :issn, :journal_title, :journal_volume, :journal_issue, :note, :orcid,
-                   :other_affiliation, :page_end, :page_start, :peer_review_status, :place_of_publication,
-                   :rights_holder, :translator, :use]
+    self.terms += [:resource_type, :abstract, :access, :bibliographic_citation, :copyright_date, :date_issued,
+                   :date_other, :dcmi_type, :doi, :edition, :extent, :funder, :geographic_subject, :issn,
+                   :journal_issue, :journal_title, :journal_volume, :note, :page_end, :page_start, :peer_review_status,
```
bbpennel: journal_issue is in a different order now, guessing we want to retain the original order with it after journal_volume

app/views/records/edit_fields/_creator.html.erb
```diff
@@ -1 +1,93 @@
-<%= f.input :creator, as: :multi_value, input_html: { class: 'form-control' }, placeholder: 'Smith, John' %>
\ No newline at end of file
+<hr>
```
lfarrell: I couldn't add an additional creator. When I click the "Add another link" nothing happens

lfarrell: Might want to add more separation and maybe a header between the people object blocks. It's a little confusing to see that multiple fields are a related block.

app/actors/hyrax/actors/general_actor.rb
smithjp: I had to add this to allow updates to nested objects to be saved.  This needs to be moved to the base actor so that it will work across all work types.  I will add a comment there.

app/views/records/edit_fields/_creator.html.erb
smithjp: It looks like I forgot to add creator and contributor to the js file.

smithjp: Ok, I'll make the separator larger and add a header for each person type.

app/renderers/hyrax/renderers/person_attribute_renderer.rb
```diff
@@ -0,0 +1,25 @@
+module Hyrax
+  module Renderers
+    class PersonAttributeRenderer < AttributeRenderer
+      private
+
+      def attribute_value_to_html(value)
+        begin
+          person = value.split(';')
```
bbpennel: I think we need to escape user input `;` characters or use a less common delimiter. If I type "Hello; How; Are; You" into the creator name in it formats each work as its own sublist entry. I know this isn't a likely name, but it gives users some unexpected control over the display.

Also, I appear to be able to do some HTML injection via the person fields, like I can bold fields or add divs.

app/views/records/edit_fields/_advisor.html.erb
```diff
@@ -1 +1,93 @@
-<%= f.input :advisor, as: :multi_value, input_html: { class: 'form-control' }, placeholder: 'Smith, John' %>
\ No newline at end of file
+<hr>
+<div id="advisor">
```
bbpennel: all of these edit_fields for person objects are going to be basically the same, aside from field name, aren't they? It seems like it would be easier to maintain/less code if we used common parameterized template that substituted "advisor" for the field name.

app/models/masters_paper.rb
smithjp: I added an advisor label field containing advisor names and made it a faceted field.

PR #197, opened 2018-11-27
Update email text

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2293


smithjp:
Should we also go ahead and include the custom workflow notification messages?

lfarrell:
Added in the workflow messages

PR #199, opened 2018-11-28
Update Hyrax to 2.4.1

lfarrell:
https://jira.lib.unc.edu/browse/HYC-257
Depends on: https://gitlab.lib.unc.edu/cdr/hyc-gems/merge_requests/6


app/views/catalog/_search_form.html.erb
```diff
@@ -0,0 +1,41 @@
+<%# [hyc-override] Remove the search form label %>
```
bbpennel: Since this is a newly overridden file, could you check in the original and then the updates in separate commits?

lfarrell: Updated with whole template as a commit

PR #201, opened 2018-11-30
Migration metadata updates

smithjp:


lib/tasks/migrate/services/metadata_parser.rb
```diff
@@ -46,25 +52,34 @@ def parse
           work_attributes['title'] = descriptive_mods.xpath('mods:titleInfo[not(@*)]/mods:title', MigrationConstants::NS).map(&:text)
           work_attributes['label'] = work_attributes['title']
           work_attributes['alternative_title'] = descriptive_mods.xpath("mods:titleInfo[@type='alternative' or @type='translated']/mods:title", MigrationConstants::NS).map(&:text)
-          work_attributes['creator'] = parse_names_from_mods(descriptive_mods, 'Creator')
-          work_attributes['contributor'] = parse_names_from_mods(descriptive_mods, 'Contributor')
-          work_attributes['advisor'] = parse_names_from_mods(descriptive_mods, 'Thesis advisor')
+          work_attributes['creators_attributes'] = parse_people_from_mods(descriptive_mods, 'Creator')
```
bbpennel: Do you think it'd be overkill to have a helper method that did all the work attribute setting per person type? Something like `populate_person_attributes('creator', @creator_label, work_attributes, descriptive_mods)`. Its a little hard to interpret all the fields here since its fairly repetitive but occasionally the number of fields varies.

app/forms/single_value_form.rb
```diff
@@ -181,28 +181,36 @@ def self.split_affiliations(affiliations)
     end
 
     def self.build_person_display(person_attrs)
-      display_text = []
-      display_text << person_attrs['name'] if !person_attrs['name'].blank?
-      display_text << "ORCID: #{person_attrs['orcid']}" if !person_attrs['orcid'].blank?
-      display_text << "Affiliation: #{split_affiliations(person_attrs['affiliation']).join(', ')}" if !person_attrs['affiliation'].blank?
-      display_text << "Other Affiliation: #{person_attrs['other_affiliation']}" if !person_attrs['other_affiliation'].blank?
-      display_text.join('||')
+      displays = []
```
bbpennel: I'm still feeling a bit unsure about the display form of the person object being stored in fedora, it will mean that changing the field labels (like if we decide "Other Affiliation" is too long, or don't want to display one of the fields anymore) will require a bulk update of the fedora records, or some weird string replacement in the UI. Is this is because we need the field to be fully formatted in Solr, but can't do this reformatting during solrizing?

lib/tasks/migrate/services/metadata_parser.rb
```diff
@@ -251,6 +271,53 @@ def parse_names_from_mods(mods, name)
...
+            person_hash[index.to_s] = { 'name' => name,
+                                        'orcid' => orcid,
+                                        'affiliation' => affiliation.join(', '),
+                                        'other_affiliation' => other_affiliation }
+          end
+
+          person_hash.blank? ? nil : person_hash
+        end
+
+        def build_person_display(people)
```
bbpennel: Aside from my worries about storing the display version, would it be possible to write a shared utility method for generate the display? Asking because it looks like this is supposed to produce the same output as `single_value_form` version, but is implemented a bit differently and does not have the splitting and joining section for Affiliation, so the outcome would differ slightly.

lib/tasks/migrate/services/ingest_service.rb
```diff
@@ -172,7 +171,13 @@ def work_record(work_attributes)
           end
 
           # Only keep attributes which apply to the given work type
-          resource.attributes = work_attributes.reject{|k,v| !resource.attributes.keys.member?(k.to_s)}
+          work_attributes.select {|k,v| k.ends_with? '_attributes'}.each do |k,v|
```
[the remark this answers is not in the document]
smithjp: The contributor attribute is automatically added to every model, but some models do not allow nested contributor attributes because the contributor attribute was not in the MAP for those models.  The next lines remove any attributes which are parsed from the mods records but are not allowed in the chosen model/work type.  Otherwise, the script fails.

app/forms/single_value_form.rb
smithjp: I'm not sure how I could modify the solrizing process in hyrax.  It would be nice to have the display information only in solr, though.

lib/tasks/migrate/services/metadata_parser.rb
smithjp: The methods are different because the form just provides the most specific department name, but the foxml files include all of the departments.  The extra department names for the form input need to be pulled from the department service, which then requires the splitting and joining for affiliations.

lib/tasks/migrate/services/metadata_parser.rb
```diff
@@ -271,6 +261,15 @@ def parse_names_from_mods(mods, type)
           name_array
         end
 
+        def get_person_info(mods, type)
+          if type == 'advisor'
+            @work_attributes[type+'s_attributes'] = parse_people_from_mods(mods, 'Thesis advisor')
```
bbpennel: I'm probably missing something, but why not pass `work_attributes` into this helper method rather than making it an instance variable everywhere? Its not being assigned a new value, only fields within it are being replaced.

lib/tasks/migrate/services/ingest_service.rb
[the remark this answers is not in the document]
bbpennel: If we switch to using the same model for every work type, will this no longer be necessary?

app/forms/single_value_form.rb
bbpennel: It seems like it happens in https://github.com/samvera/active_fedora/blob/master/lib/active_fedora/rdf/indexing_service.rb
It might be possible to update https://github.com/samvera/active_fedora/blob/master/lib/active_fedora/rdf/indexing_service.rb#L34 to detect and format Person object fields

lib/tasks/migrate/services/metadata_parser.rb
bbpennel: Okay, thank you for the explanation. Could you add a comment stating this difference, just so you won't have to reexplain it everytime someone (like me :) ) looks at the method in the future.

PR #205, opened 2018-12-11
Remove batch deletion

lfarrell:
https://jira.lib.unc.edu/browse/HYC-375


app/views/hyrax/my/works/_batch_actions.html.erb
```diff
@@ -0,0 +1,16 @@
+<%# [hyc-override] Overriding partial in hyrax gem to remove batch delete link %>
+<div class="batch-info">
+  <%= render 'hyrax/dashboard/collections/form_for_select_collection', user_collections: @user_collections %>
+
+  <div class="batch-toggle">
+    <% session[:batch_edit_state] = "on" %>
+    <% if Flipflop.batch_upload? %>
+      <div class="button_to-inline">
+        <%= button_to t('hyrax.dashboard.my.action.edit_selected'), hyrax.edit_batch_edits_path, method: :get, class: "btn btn-update btn-primary hidden submits-batches", data: { behavior: 'batch-edit' }, id: 'batch-edit' %>
```
bbpennel: This PR appears to remove the Delete button but not the "Edit Selected" button. We should probably remove that too since this PR results in clicking it giving a "Batch operations are not allowed" message

PR #209, opened 2018-12-13
Make sure required people object fields are required

lfarrell:
https://jira.lib.unc.edu/browse/HYC-368

### DIFF
--- a/app/assets/javascripts/hyrax/save_work/required_fields.es6
+++ b/app/assets/javascripts/hyrax/save_work/required_fields.es6
@@ -1,0 +1,23 @@
+export class RequiredFields {
+    // Monitors the form and runs the callback if any of the required fields change
+    constructor(form, callback) {
+        this.form = form
+        this.callback = callback
+        this.reload()
+    }
+
+    get areComplete() {
+        return this.requiredFields.filter((n, elem) => { return this.isValuePresent(elem) } ).length === 0
+    }
+
+    isValuePresent(elem) {
+        return ($(elem).val() === null) || ($(elem).val().length < 1)
+    }
+
+    // Reassign requiredFields because fields may have been added or removed.
+    reload() {
+        // ":input" matches all input, select or textarea fields.
+        this.requiredFields = this.form.find(':input[required]')
+        this.requiredFields.change(this.callback)
+    }
+}
--- a/app/assets/javascripts/hyrax/save_work/required_fields.es6
+++ b/app/assets/javascripts/hyrax/save_work/required_fields.es6
@@ -1,4 +1,4 @@
-// [hyc-override] Override to allow form to be submitted with hidden, empty, cloning people object fields
+// [hyc-override] Override to allow form to be submitted with hidden, empty, cloning fields
 export class RequiredFields {
     // Monitors the form and runs the callback if any of the required fields change
     constructor(form, callback) {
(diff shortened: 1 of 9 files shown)


app/assets/javascripts/unc/unc_custom.js
```diff
@@ -53,11 +53,19 @@ $(function() {
         $('a.additional-fields').attr('aria-expanded', false);
     }
 
+    // Remove hidden, cloning people elements. They prevent forms from submitting as they have required fields
+    function removeCloning() {
```
bbpennel: Just curious, why is removing these a new requirement? The `cloning` creation isn't new to this PR

app/assets/javascripts/hyrax/save_work/required_fields.es6
```diff
@@ -0,0 +1,33 @@
...
+export class RequiredFields {
+    // Monitors the form and runs the callback if any of the required fields change
+    constructor(form, callback) {
+        this.form = form
+        this.callback = callback
+        this.reload()
+    }
+
+    get areComplete() {
+        return this.requiredFields.filter((n, elem) => { return this.isValuePresent(elem) } ).length === 0
```
bbpennel: I wonder what would happen if during this check it made sure the element wasn't a `cloning` one, versus removing all the cloning elements?

app/forms/hyrax/honors_thesis_form.rb
```diff
@@ -12,7 +12,7 @@ class HonorsThesisForm < ::SingleValueForm
 
     self.terms -= [:based_near, :contributor, :description, :identifier, :publisher, :source]
 
-    self.required_fields = [:title, :creator, :abstract, :advisor, :degree, :award, :date_issued,
+    self.required_fields = [:title, :creator, :abstract, :advisor, :affiliation, :degree, :award, :date_issued,
```
bbpennel: If i'm understanding correctly, `affilation` is now a required field, but is not present in the form? Is that where the need to remove "cloning" elements on submit comes from?

app/views/records/edit_fields/_person.html.erb
```diff
@@ -65,6 +67,7 @@
             <%= person.input :affiliation, as: :select,
                               collection: DepartmentsService.select_all_options.sort,
                               include_blank: true,
+                              required: f.object.required?(:affiliation),
```
bbpennel: there's not a way to make the person object field itself restricted is there? like checking for `f.object.required?(:creator_affiliation)` or something? I know its on a sub-object so it might not be directly referencable to the here.

app/forms/hyrax/honors_thesis_form.rb
lfarrell: yes, as it's checking if the field is in the required fields block

app/views/records/edit_fields/_person.html.erb
lfarrell: I don't think so. I believe I tried that at one point. I'll double check.

app/assets/javascripts/hyrax/save_work/required_fields.es6
lfarrell: This just checks whether the submit button should be enabled and the check mark shown. I tried that. :)

app/assets/javascripts/unc/unc_custom.js
lfarrell: Since they're now required. In the old setup, any subsequent creator, whatever had no fields that were required and didn't show the "required" field. They didn't save anything if the name was missing, though.

app/views/records/edit_fields/_person.html.erb
lfarrell: Nope, seems to ignore it

app/assets/javascripts/unc/unc_custom.js
lfarrell: Fixed the submit button and valid form checks, though it semi-parallels the default hyrax stuff. It should be noted, though that cloning a required field in hyrax doesn't make it required, just the first instance of the field.

PR #211, opened 2018-12-17
Allow embargo updates for forms with required single valued fields

lfarrell:
https://jira.lib.unc.edu/browse/HYC-384

### DIFF
--- a/app/forms/single_value_form.rb
+++ b/app/forms/single_value_form.rb
@@ -47,17 +47,11 @@ class SingleValueForm < Hyrax::Forms::WorkForm
     is_new_model = !form_params.has_key?(:permissions_attributes)
     if is_new_model
       default_term_values.each do |field, values|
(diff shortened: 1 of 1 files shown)


app/forms/single_value_form.rb
```diff
@@ -76,7 +70,13 @@ def self.model_attributes(form_params)
     end
 
     if attrs.key?(:rights_statement) && !attrs[:rights_statement].blank?
-      attrs[:rights_statement_label] = CdrRightsStatementsService.label(attrs[:rights_statement])
+      begin
+        Rails.logger.info "Boogie: #{attrs[:rights_statement]}"
```
bbpennel: I think you will want to boogie and doogie these comments out of here

lfarrell: Ha, at least there's no cursing. Should be fixed now. :)

PR #225, opened 2019-01-04
Change work type human readable name

lfarrell:
https://jira.lib.unc.edu/browse/HYC-389


smithjp:
It is also the only plural option in the child work selector.  We may want to have a separate label for the modals, since that is the only area I am seeing the work types pluralized.

bbpennel:
It looks like we want to display "Poster, Presentation or Paper" at the top of the "Add new work" page. That may resolve @smithjp's observation as well.

lfarrell:
updated

PR #230, opened 2019-01-07
Adjust multi-select box size

lfarrell:
https://jira.lib.unc.edu/browse/HYC-393

### DIFF
--- a/app/assets/stylesheets/unc_custom.css.scss
+++ b/app/assets/stylesheets/unc_custom.css.scss
@@ -340,6 +340,11 @@ div.unc-citation-modal {
   width: 100%;
 }
 
+select[multiple],
+select[size] {
+  min-height: 115px;
+}
+
 /* Hyrax override */
 #content_block_page {
   margin: 0 15px;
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/unc_custom.css.scss
```diff
@@ -340,6 +340,11 @@ div.unc-citation-modal {
   width: 100%;
 }
 
+select[multiple],
```
bbpennel: It looks like you can change the form element with
`input_html: { class: 'form-control', multiple: true, size: 6 }`
So that it wouldn't impact all other multi selects.

I can see why changing the css would be appealing since we currently have 6 variations on the resource type box. They appear to all be identical except for the default selected value. It would be good to refactor that to be less repetitive. My guess is it could be rewritten to be a single field declaration and put the logic that determines the default value in a block before it.
```
if f.object.resource_type[0].blank?
  // case statement to set default_selected
else
  default_selected = f.object.resource_type[0]
end
```
or something like that

lfarrell: Updated

bbpennel:
The most recent changes were causing the resource type to be lost when going to the edit form. I've pushed a commit that refactors the field a little and fixes that issue.

PR #233, opened 2019-01-10
Numbering People Objects

lfarrell:
https://jira.lib.unc.edu/browse/HYC-401

### DIFF
--- a/app/views/records/edit_fields/_person.html.erb
+++ b/app/views/records/edit_fields/_person.html.erb
@@ -34,7 +34,7 @@
 
       <div class="person col-sm-12 <%= person_term %>-other-affiliation">
         <%= person.input :other_affiliation,
-                         label: "Other Affiliation (#{person_term.split('_').map(&:capitalize).join(' ')} #1)",
+                         label: "Additional affiliation (#{person_term.split('_').map(&:capitalize).join(' ')} #1)",
                          input_html: { class: 'form-control',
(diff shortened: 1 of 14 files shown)


app/views/records/edit_fields/_person.html.erb
```diff
@@ -1,36 +1,41 @@
...
 <div class="form-group multi_value">
   <div id="<%= person_term %>">
-  <h3 align="center"><%= person_term.split('_').map(&:capitalize).join(' ') %></h3>
   <%= f.fields_for person_key do |person| %>
-    <div class="<%= person_term %> row well field-wrapper">
+    <div class="<%= person_term %> row well field-wrapper person-object">
       <div class="person col-sm-12 <%= person_term %>-name">
         <%= person.input :name,
-                          input_html: { placeholder: 'Smith, John',
+                         label: "#{person_term.split('_').map(&:capitalize).join(' ')} #1",
```
bbpennel: Seems like you could compute `Creator #1` once to store into a variable and reuse it for all the fields? Maybe even use that same variable for both the top and follow up entries since apparently they all show up as #1. I guess the javascript changes all the indexes to the correct number.

lfarrell: Not sure I quite get what you're saying. When you remove a person it could be anywhere in the ordering. So, you'd have to loop through them all to update. I'm guessing your talking about initial row creation numbering, not row deletion?

PR #234, opened 2019-01-11
HYC-332 - Trigger longleaf via external script

bbpennel:
https://jira.lib.unc.edu/browse/HYC-332


bbpennel:
@lfarrell Its calling a command which does the actual call to longleaf. When the command is called, it is passed arguments to pass along to Longleaf, which includes the file to register. This was so that more complicated stuff could optionally happen when calling longleaf, like ssh connections, retries, sending the logging weird places, etc.

PR #249, opened 2019-01-22
Add methodology form field

lfarrell:
https://jira.lib.unc.edu/browse/HYC-433


bbpennel:
Could you add the help text for this field, "Describe the methods used to generate the dataset, or how the dataset may be properly reused"?

PR #252, opened 2019-01-24
Delete people objects when editing

lfarrell:
https://jira.lib.unc.edu/browse/HYC-425


app/assets/javascripts/person_objects.js
```diff
@@ -42,64 +46,84 @@ function attach_add_person_listeners(selector){
...
-        var self = $(this);
-        self.prop('name', self.prop('name').replace(regex, index))
-            .attr('id', self.attr('id').replace(regex, index));
-    });
-    
-    // Update label
-    var $label = new_row.find("label[for$='" + attr + "']");
+    ['name', 'orcid', 'affiliation', 'other_affiliation'].forEach(function(attr) {
+        // Update input
+        if (!is_deletion) {
```
bbpennel: Move this out to just happen during row creation

app/assets/javascripts/person_objects.js
```diff
@@ -42,64 +46,84 @@ function attach_add_person_listeners(selector){
...
 }
 
-function updateFields(row, current_index) {
-    ['name', 'orcid', 'affiliation', 'other_affiliation'].forEach(function(field) {
-        updateRow(row, field, current_index);
+function updateAllRows(row_selector, is_deletion, label_index) {
+    $(row_selector).not(':hidden').each(function(row_index) {
+        var self = $(this);
+        var offset = label_index - row_index;
+        updateFields(self, is_deletion, row_index, label_index - offset);
```
bbpennel: see if you can just pas in `row_index + 1` instead of `label_index - offset`

app/assets/javascripts/person_objects.js
```diff
@@ -3,11 +3,11 @@ $(document).on('turbolinks:load', function () {
         'reviewer', 'translator'];
     people.forEach(function(person) {
         attach_add_person_listeners(person);
```
bbpennel: You could replace the hidden fields being added in the erb by doing a count of the number of rows per person type at page load, and storing that value somewhere in javascript land (like a `.data` field on the `$("#creator")` object, or some kind of array, whatever).

app/assets/javascripts/person_objects.js
```diff
@@ -42,64 +46,84 @@ function attach_add_person_listeners(selector){
         $('div#'+selector).append($new_row);
 
         current_index++;
+        current_label_index++;
         index_selector.val(current_index);
+        label_index_selector.val(current_label_index);
```
bbpennel: You could count the number of people for this type at the time of adding, then use that in place of tracking label_index_selector. `label_index_selector` is probably only needed for adding new people, since during an updateAllRows it will just start at 0 and go up. That may simplify things a little.

app/assets/javascripts/person_objects.js
```diff
@@ -3,11 +3,11 @@ $(document).on('turbolinks:load', function () {
         'reviewer', 'translator'];
     people.forEach(function(person) {
         attach_add_person_listeners(person);
-        updateAllRows('.'+ person + '.row');
+        updateAllRows('.'+ person + '.row', false, $('#index-label-' + person).val());
```
bbpennel: updateAllRows is in camel case, and some of the other functions are too, while some others are _ separated. Might as well be consistent.

app/assets/javascripts/person_objects.js
```diff
@@ -1,134 +1,182 @@
...
+        var remove_selector = '.remove-' + person_type;
+        var cloning_row = '#' + person_type + '-cloning_row';
 
-    $(add_selector).on('click', function(event) {
-        // stop page from reloading
-        event.preventDefault();
+        $person_row.find(remove_selector).on('click', function (event) {
+            event.preventDefault();
+        
+            var $self = $(this);
```
lfarrell: I don't think this is getting used.

bbpennel: ah you're right, the need for it came and went a few times

PR #254, opened 2019-01-24
Create deletion role for withdrawing/deleting records

lfarrell:
https://jira.lib.unc.edu/browse/HYC-442


bbpennel:
@lfarrell okay, sounds like ir-qa is usable for this test now

PR #264, opened 2019-02-05
Box-c to hy-c redirects

lfarrell:
https://jira.lib.unc.edu/browse/BXC-2319
Related to https://gitlab.lib.unc.edu/cdr/vagrant-rails/merge_requests/54

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -26,18 +26,23 @@ class ApplicationController < ActionController::Base
   # Can be removed if we no longer need redirects
   def check_redirect
     # Base redirect for Box-C uuid links
-    full_path = request.url
-    path_parts = request.url.split('uuid:')
+    full_path = request.original_fullpath
     request_host = "#{request.protocol}#{request.host}"
 
     if request_host =~ /localhost/
       request_host = "#{request_host}:#{request.port}"
     end
 
+    uuid = full_path[/uuid:([a-f0-9\-]+)/, 1]
+
     # Base redirect for Hy-C uuid links
-    if path_parts.length > 1
-      uuid = path_parts[1]
-      redirect_uuids = File.read(Rails.root.join('lib', 'redirects', 'redirect_uuids.csv'))
+    unless uuid.nil?
+      if ENV.has_key?('REDIRECTS_FILE_PATH') && File.exist?(ENV['REDIRECTS_FILE_PATH'])
+        redirect_uuids = File.read(ENV['REDIRECTS_FILE_PATH'])
+      else
+        redirect_uuids = File.read(Rails.root.join('lib', 'redirects', 'redirect_uuids.csv'))
+      end
+
       csv = CSV.parse(redirect_uuids, headers: true)
       redirect_path = csv.find { |row| row['uuid'] == uuid }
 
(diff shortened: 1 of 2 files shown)


app/controllers/application_controller.rb
```diff
@@ -22,6 +23,52 @@ def check_read_only
...
+      redirect_path = csv.find { |row| row['uuid'] == uuid }
+
+      # Should correctly redirect record, indexablecontent (download) paths
+      if redirect_path # Redirect to Hy-C
+        updated_path = "#{request_host}/concern/#{redirect_path['new_path']}"
+        Rails.logger.info "In hy-c uuid redirect match: #{updated_path}"
+        redirect_to updated_path, status: :moved_permanently
+      elsif full_path =~ /search.*?uuid/ # All Box-C searches with uuids should go to the 404 page
+        updated_path = "#{request_host}/concern/404"
+        Rails.logger.info "Is box-c search with uuid: #{updated_path}"
```
bbpennel: I don't think "updated_path" is what we want to log here since that is the error page at this point.

app/controllers/application_controller.rb
```diff
@@ -22,6 +23,52 @@ def check_read_only
     )
   end
 
+  # Can be removed if we no longer need redirects
+  def check_redirect
+    # Base redirect for Box-C uuid links
+    full_path = request.url
```
bbpennel: Optional, but it might make sense to use `request.original_fullpath` instead. That would allow you to use `starts_with?` for a lot of the `if` statements below rather than searching for keyswords anywhere in the url. You only need the full url when performing the substitution at line 54

app/controllers/application_controller.rb
```diff
@@ -46,27 +51,31 @@ def check_redirect
         updated_path = "#{request_host}/concern/#{redirect_path['new_path']}"
         Rails.logger.info "In hy-c uuid redirect match: #{updated_path}"
         redirect_to updated_path, status: :moved_permanently
-      elsif full_path =~ /search.*?uuid/ # All Box-C searches with uuids should go to the 404 page
+      elsif full_path.starts_with?('/search') # All Box-C searches with uuids should go to the 404 page
```
bbpennel: Sorry i missed this the first time around, could you add '/list' and '/listContent' here?

app/controllers/application_controller.rb
```diff
@@ -22,6 +23,61 @@ def check_read_only
...
+      redirect_path = csv.find { |row| row['uuid'] == uuid }
+
+      # Should correctly redirect record, indexablecontent (download) paths
+      if redirect_path # Redirect to Hy-C
+        updated_path = "#{request_host}/concern/#{redirect_path['new_path']}"
+        Rails.logger.info "In hy-c uuid redirect match: #{updated_path}"
+        redirect_to updated_path, status: :moved_permanently
+      elsif full_path.starts_with?('/search') # All Box-C searches with uuids should go to the 404 page
+        updated_path = "#{request_host}/concern/404"
+        Rails.logger.info "Box-c search with uuid: #{request_host}/concern/404"
```
bbpennel: Seems like it would be more helpful if it logged the original path rather than where its being forwarded to. Maybe change the message to `"Forwarding Box-c search to 404, user requested #{full_path}"`

app/controllers/application_controller.rb
```diff
@@ -22,6 +23,61 @@ def check_read_only
...
+
+    if request_host =~ /localhost/
+      request_host = "#{request_host}:#{request.port}"
+    end
+
+    uuid = full_path[/uuid:([a-f0-9\-]+)/, 1]
+
+    # Base redirect for Hy-C uuid links
+    unless uuid.nil?
+      if ENV.has_key?('REDIRECTS_FILE_PATH') && File.exist?(ENV['REDIRECTS_FILE_PATH'])
```
bbpennel: In the ansible PR the env variable is REDIRECT_FILE_PATH, so i think you need to change the key over here to match it (remove the s)

PR #265, opened 2019-02-05
HYC446 - Cleanup uploaded files

bbpennel:
Deletes uploaded files after derivative generation finishes, as this is the last step in the file_actor workflow

https://jira.lib.unc.edu/browse/HYC-446


spec/jobs/create_derivatives_job_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+
+RSpec.describe CreateDerivativesJob do
+  around do |example|
+    ffmpeg_enabled = Hyrax.config.enable_ffmpeg
+    Hyrax.config.enable_ffmpeg = true
+    example.run
+    Hyrax.config.enable_ffmpeg = ffmpeg_enabled
+  end
+
+  context "with an audio file" do
```
lfarrell: Is there an audio file being tested? I'm seeing a PDF

bbpennel:
@smithjp I think that is the case. AttachFilesToWorkJob creates the FileSetActor which creates the FileActor which ingests the file and adds it to the fileset before it queues up the CharacterizeJob which eventually leads to the triggering of derivative jobs and cleanup. So the File and FileSet must have been created/ingested before cleanup happens.

PR #266, opened 2019-02-06
Add status monitor for sidkiq

lfarrell:
Related to https://gitlab.lib.unc.edu/cdr/hyc-gems/merge_requests/8


lfarrell:
Had to disable clamav

bbpennel:
I am guessing this requires an update to the hyc-gems project?

smithjp:
clamav still needs to be in the gemfile and the hyc-gems project so it can be deployed on ir-test and ir-qa.

PR #272, opened 2019-02-08
Read settings from an ENV variable

lfarrell:
https://jira.lib.unc.edu/browse/HYC-466

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -20,7 +20,7 @@ module Hyrax
 
       YAML.load(File.open(env_file)).each do |key, value|
         ENV[key.to_s] = value
-      end
+      end if File.exists?(env_file)
     end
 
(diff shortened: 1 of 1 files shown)


config/application.rb
```diff
@@ -12,13 +12,18 @@ class Application < Rails::Application
...
-       end if File.exists?(env_file)
+      if ENV.has_key?('LOCAL_ENV_PATH')
+        env_file = ENV['LOCAL_ENV_PATH'].to_s
+      else
+        env_file = File.join(Rails.root, 'config', 'local_env.yml')
+      end
+
+      YAML.load(File.open(env_file)).each do |key, value|
+        ENV[key.to_s] = value
+      end if File.exists?(env_file)
```
bbpennel: looking at this now, i think we should remove the if here. That way if a non-existent local_env.yml was specified the error will be a failure to open the file, rather than missing properties elsewhere.

PR #282, opened 2019-02-13
Perform lookup of geonames urls to get human readable text value

lfarrell:
https://jira.lib.unc.edu/browse/HYC-482


lfarrell:
Does the reverse of what we want, so closing.

PR #285, opened 2019-02-15
Honors workflow update

smithjp:
fixing problems created by merging the wrong PR

### DIFF
--- a/config/workflows/honors_thesis_deposit_workflow.json
+++ b/config/workflows/honors_thesis_deposit_workflow.json
@@ -13,11 +13,6 @@
                     "notifications": [
                         {
                             "notification_type": "email",
-                            "name": "Hyrax::Workflow::PendingReviewNotification",
-                            "to": ["approving"]
-                        },
-                        {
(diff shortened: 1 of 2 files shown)


bbpennel:
It looks like the last update on this branch was:
```
julesies [10:19 AM]
ok we tested all workflows. all work as expected EXCEPT Default workflows is sending manager a deposited work notice when it’s not supposed to.

we were not sure if those fixes are both on QA or not?

jennifer [10:21 AM]
It looks like the default workflow updates are currently on ir-qa.  Only the depositor should be getting deposit notifications.

julesies [10:22 AM]
ok so that is a bug i guess.
everything else is working correctly!

jennifer [10:25 AM]
I'll double check the notifiable contexts on ir-qa to make sure that email configurations are being cleared out.
```

What stage is this work in at this point?

smithjp:
I was looking at the notification problems yesterday, and it appears to be a larger problem with the way sipity works and not something introduced with these changes.  I will create a new ticket for fixing the new workflow notification problems.

PR #289, opened 2019-02-18
Allow multiple date formats

lfarrell:
https://jira.lib.unc.edu/browse/HYC-489


app/assets/javascripts/hyrax/save_work/required_fields.es6
```diff
@@ -21,13 +23,41 @@ export class RequiredFields {
         }
 
         let selectorValue = selector.val();
-        return (selectorValue === null) || (selectorValue.length < 1);
+        let isSelected = (selectorValue === null) || (selectorValue.length < 1);
+
+        if (/date_created/.test(selector.prop('name')) || /date_issued/.test(selector.prop('name'))) {
+            // (jan(uary)?|feb(ruary)?|mar(ch)?|apr(il)?|may|june|july|aug(ust)?|sept(ember)?|oct(ober)?|nov(ember)?|dec(ember)?|circa|spring|summer|fall|winter|autumn)?\s{0,}\d{4}s? matches Spring 2000, Aug 2000, July 2000, 2000, 2000s, circa 2000
```
bbpennel: Rather than repeating the regex in the comment, maybe move the comments describing what it does down into the regex array below?

app/assets/javascripts/hyrax/save_work/required_fields.es6
```diff
@@ -21,13 +23,41 @@ export class RequiredFields {
...
+            // \d{4}(\\|\/|-)\d{2}(\\|\/|-)\d{2} matches 2000-01-01, 2000/01/01, 2000\01\01
+            // \d{2}(\\|\/|-)\d{2}(\\|\/|-)\d{4} matches 01-01-2000, 01/01/2000, 01\01\2000
+            // \d{4}(\s{0,}to\s{0,}|-)\d{4} matches 2000 to 2010, 2000-2010
+            // (jan(uary)?|feb(ruary)?|mar(ch)?|apr(il)?|may|june|july|aug(ust)?|sept(ember)?|oct(ober)?|nov(ember)?|dec(ember)?)\s+\d{1,2}[a-zA-Z]{0,2},?\s{0,}\d{4} matches July 1st 2000, July 1st, 2000, July 1 2000, July 1, 2000
+            // unknown matches Unknown, unknown
+            // ^$ matches empty string
+            let regex = [
+                '^((jan(uary)?|feb(ruary)?|mar(ch)?|apr(il)?|may|june|july|aug(ust)?|sept(ember)?|oct(ober)?|nov(ember)?|dec(ember)?|circa|spring|summer|fall|winter|autumn)?\\s{0,}\\d{4}s?',
```
bbpennel: Please replace all the `{0,}` with `*`

app/assets/javascripts/hyrax/save_work/required_fields.es6
```diff
@@ -21,13 +23,41 @@ export class RequiredFields {
...
+                '\\d{4}(\\\\|\\/|-)\\d{2}(\\\\|\\/|-)\\d{2}',
+                '\\d{2}(\\\\|\\/|-)\\d{2}(\\\\|\\/|-)\\d{4}',
+                '\\d{4}(\\s{0,}to\\s{0,}|-)\\d{4}',
+                '(jan(uary)?|feb(ruary)?|mar(ch)?|apr(il)?|may|june|july|aug(ust)?|sept(ember)?|oct(ober)?|nov(ember)?|dec(ember)?)\\s+\\d{1,2}[a-zA-Z]{0,2},?\\s{0,}\\d{4}',
```
bbpennel: `\\d{1,2}[a-zA-Z]{0,2}` is intended to match 1st, etc. I'd suggest putting in the specific allowed values to avoid "March 30uh, 2019", so:
`\\d{1,2}(st|nd|rd|th)?`

app/assets/javascripts/hyrax/save_work/required_fields.es6
```diff
@@ -21,13 +23,41 @@ export class RequiredFields {
...
+                'unknown|^$)$'
```
bbpennel: the regex already begins and ends with ^$, guessing you can switch to ` 'unknown)?$'

app/assets/javascripts/hyrax/save_work/required_fields.es6
```diff
@@ -21,13 +23,41 @@ export class RequiredFields {
...
+            // (jan(uary)?|feb(ruary)?|mar(ch)?|apr(il)?|may|june|july|aug(ust)?|sept(ember)?|oct(ober)?|nov(ember)?|dec(ember)?|circa|spring|summer|fall|winter|autumn)?\s{0,}\d{4}s? matches Spring 2000, Aug 2000, July 2000, 2000, 2000s, circa 2000
+            // \d{4}(\\|\/|-)\d{2}(\\|\/|-)\d{2} matches 2000-01-01, 2000/01/01, 2000\01\01
+            // \d{2}(\\|\/|-)\d{2}(\\|\/|-)\d{4} matches 01-01-2000, 01/01/2000, 01\01\2000
+            // \d{4}(\s{0,}to\s{0,}|-)\d{4} matches 2000 to 2010, 2000-2010
+            // (jan(uary)?|feb(ruary)?|mar(ch)?|apr(il)?|may|june|july|aug(ust)?|sept(ember)?|oct(ober)?|nov(ember)?|dec(ember)?)\s+\d{1,2}[a-zA-Z]{0,2},?\s{0,}\d{4} matches July 1st 2000, July 1st, 2000, July 1 2000, July 1, 2000
+            // unknown matches Unknown, unknown
+            // ^$ matches empty string
+            let regex = [
+                '^((jan(uary)?|feb(ruary)?|mar(ch)?|apr(il)?|may|june|july|aug(ust)?|sept(ember)?|oct(ober)?|nov(ember)?|dec(ember)?|circa|spring|summer|fall|winter|autumn)?\\s{0,}\\d{4}s?',
+                '\\d{4}(\\\\|\\/|-)\\d{2}(\\\\|\\/|-)\\d{2}',
```
bbpennel: Does anyone really use \ delimiters for date fields? gross, looks very windows. I don't see them in the "Separators of the components" section of https://en.wikipedia.org/wiki/Date_format_by_country not that i think we need to support spaces and dots as separators. Anyways, I'd be okay removing the `\`'s.

app/lib/hyc/validators/date_validator.rb
```diff
@@ -0,0 +1,39 @@
+module Hyc::Validators
```
bbpennel: It seems like this PR is validating the date fields, but my understanding is that it is a requirement for the fields to be stored as dates in EDTF format rather than the original human syntax. You're welcome to check with julie/anna/rebekah about that though.

app/assets/javascripts/hyrax/save_work/required_fields.es6
lfarrell: Testing in the console it doesn't let through an empty string unless the ^$ bit is there as specific a option.

app/forms/hyc/edtf_convert.rb
```diff
@@ -0,0 +1,64 @@
...
+    def convertToEdtf(field)
+      case field
+      when /^\d{4}s$/ # 1900s, 1980s
+        normalized_string = (field[3] == 0 && field[4] == 0) ? "#{field[0...-3]}xx" : "#{field[0...-2]}x"
+      when /^(\d{4}-\d{2}-\d{2}|\d{4})$/ # 2000-01-01 or 2000
+        normalized_string = field
+      when /^\d{2}-\d{2}-\d{4}$/ # 01-01-2000
+        date_parts = split_date(field, '-')
+        normalized_string = "#{date_parts[2]}-#{date_parts[0]}-#{date_parts[1]}"
+      when /^\d{4}(\s+to\s+|-)\d{4}$/ # 1990 to 1995
```
bbpennel: there could be spacing around the `-` as well.

Also, it looks like you can get the match results from the regex in the when statement:
https://stackoverflow.com/questions/6803647/how-to-write-a-ruby-switch-statement-case-when-with-regex-and-backreferences

So if the regex were changed to `/^(\d{4})(\s+to\s+|\s*-\s*)(\d{4})$/
Then I believe you could skip the split_date call and return `"#{$1}/#{$3}"`

This could be applied to a number of other cases here.

app/forms/hyc/edtf_convert.rb
```diff
@@ -0,0 +1,64 @@
...
+        date_parts = split_date(field, /(\s+to\s+|-)/)
+        normalized_string = "#{date_parts[0]}/#{date_parts[1]}"
+      when /'^(jan(uary)?|feb(ruary)?|mar(ch)?|apr(il)?|may|june|july|aug(ust)?|sept(ember)?|oct(ober)?|nov(ember)?|dec(ember)?|spring|summer|fall|winter|autumn)\s*\d{4}$/i # Spring 2000, Aug 2000, July 2000
```
bbpennel: You could create most of this regex by joining the set of keys from `@month_season` so you wouldn't need to repeat it here. That would also mean you wouldn't need all the `()?` sections and could reasonably use matching groups here to get the month/season and year when composing the normalized string.

app/forms/hyc/edtf_convert.rb
```diff
@@ -0,0 +1,64 @@
...
+        september: '09', sept: '09',
+        october: '10', oct: '10',
+        november: '11', nov: '11',
+        december: '12', dec: '12',
+        spring: '21', summer: '22',
+        fall: '23', autumn: '23', winter: '24'
+    }
+
+    def convertToEdtf(field)
+      case field
```
bbpennel: This is looking good. One option to defer some of the complexity to chronic (since it looks like we are including it anyways on another branch) would be to have this `case` catch the special cases (1900s, 1990-1995, Spring/August 2010, 2010, circa, unknown, and empty) and then send everything else to Chronic and format into yyyy-mm-dd if the result isn't nil. Honestly this would only cut out 2.5 of the cases though, so I'll leave it up to your judgment.

app/forms/single_value_form.rb
```diff
@@ -85,12 +86,34 @@ def self.model_attributes(form_params)
       end
     end
 
-    attrs
-  end
+    # Convert dates from strings to EDTF format
+    if attrs.key?(:date_created) && !attrs[:date_created].blank?
```
bbpennel: Aside from the field name, this `if` and the next one are the same, so it would make sense to break it into a well named helper method.

Also, there isn't any way to do this in the model is there? I'm just a little worried about this file getting filled up with special cases, but if not this is fine.

app/forms/hyc/edtf_convert.rb
lfarrell: Updated. Don't need the split at all if the capture groups are setup correctly

app/forms/hyc/edtf_convert.rb
lfarrell: Decided to leave out Chronic. It seemed to add some weirdness as it's trying to convert it to a Ruby date instead of checking the format. Also, Chronic isn't really maintained anymore

app/forms/single_value_form.rb
lfarrell: Added a method in the EdtfConvert module to perform the updates

app/forms/hyc/edtf_convert.rb
lfarrell: Good call. Done

app/forms/hyc/edtf_convert.rb
```diff
@@ -0,0 +1,92 @@
...
+
+    private
+
+    def convert_to_edtf(field)
+      case field
+      when /^\d{4}s$/ # 1900s, 1980s
+        normalized_string = (field[2].to_i == 0 && field[3].to_i == 0) ? "#{field[0...-3]}xx" : "#{field[0...-2]}x"
+      when /^(\d{4}-\d{2}-\d{2}|\d{4})$/ # 2000-01-01 or 2000
+        normalized_string = field
+      when /^(\d{2})-(\d{2})-(\d{4})$/ # 01-01-2000
```
bbpennel: This case should allow for `01/01/2000` format as well

app/forms/single_value_form.rb
```diff
@@ -85,30 +86,33 @@ def self.model_attributes(form_params)
...
+    # Convert dates from human readable strings to EDTF format
+    edtf_form_update(attrs, :date_created)
+    edtf_form_update(attrs, :date_issued)
+
     attrs
   end
 
+  private
 
+  def initialize_default_term_values(model)
```
bbpennel: The three of us should come to an agreement on whether or not to indent after a `private` declaration. For now, could you revert this method since the only changes are indentation?

lfarrell: It gives a style warning in RubyMine to indent. If we ever use rubocop for Hy-C we might have a lot to fix :)

app/forms/hyc/edtf_convert.rb
```diff
@@ -18,42 +19,33 @@ def edtf_form_update(attrs, field)
       attrs
     end
 
-    private
-
-    def convert_to_edtf(field)
+    def self.convert_from_edtf(field)
```
bbpennel: This appears to work, but it might be better to rely on humanize for going this direction, since we already have it as a dependency, and i think that is what is being used to format the date in the record view
```
d = Date.edtf(field)
d.humanize
```
It should return nil if it can't parse

PR #293, opened 2019-02-19
Add honors concentration field

lfarrell:
https://jira.lib.unc.edu/browse/HYC-504. 

Only in honors thesis. Can't be added to general form as it uses the same predicate as "Academic Concentration"


bbpennel:
merge conflicts, sorry :(

PR #301, opened 2019-02-26
Migrate dates and terms of address with names

smithjp:
https://jira.lib.unc.edu/browse/HYC-464


lfarrell:
This isn't being merged into master?

PR #303, opened 2019-02-26
Store Geonames URIs

lfarrell:
https://jira.lib.unc.edu/browse/HYC-501
Switch to based_near field from geographic_subject field


smithjp:
I'm getting an error when trying to migrate a record with a based near value.  I added based near information to the test metadata file in a new branch, `based-near-migration`.

PR #307, opened 2019-03-04
Add file extensions

smithjp:
https://jira.lib.unc.edu/browse/HYC-492


lib/tasks/migrate/services/ingest_service.rb
```diff
@@ -160,13 +172,26 @@ def ingest_records
...
+        else
+          extension = nil
+          puts "[#{Time.now.to_s}] #{parent.id} expected mime_type: #{resource[:mime_type]}"
+        end
+
+        # Update file title
+        if resource['title'].first.match?(/.*#{extension}\z/)
+          extension = ''
+        end
+        resource['title'] = resource['title'].map{|title| title.split(' ').join('_')+extension}
```
bbpennel: I'd like to double check that people are okay with us changing the stored file title, versus adding the extension in on retrieval

lib/tasks/migration/mime_types.csv
```diff
@@ -0,0 +1,150 @@
+type,extension
```
bbpennel: Given the addition of mimetype.yml, is this file still needed?

smithjp: no, I will remove it

PR #311, opened 2019-03-07
Some public works give authorization errors

lfarrell:
https://jira.lib.unc.edu/browse/HYC-521
https://jira.lib.unc.edu/browse/HYC-523


bbpennel:
Good work! That's a finicky difference, but makes sense.

PR #312, opened 2019-03-08
Workflow notifications

smithjp:
https://jira.lib.unc.edu/browse/HYC-516


smithjp:
We may want to squash this one.

app/services/hyrax/workflow/deposited_manager_notification.rb
```diff
@@ -0,0 +1,19 @@
+module Hyrax
+  module Workflow
+    class DepositedManagerNotification < AbstractNotification
```
bbpennel: It would be helpful to add a comment to this class to explain what is different about it versus the default notification manager.

PR #316, opened 2019-03-08
Hyc 498 primary objects

smithjp:
https://jira.lib.unc.edu/browse/HYC-498
* migrate primary objects as filesets in hyrax
* example: https://ir-test.lib.unc.edu/concern/journals/dv13zv83d?locale=en

### DIFF
--- a/lib/tasks/migrate/services/child_work_parser.rb
+++ b/lib/tasks/migrate/services/child_work_parser.rb
@@ -1,0 +1,75 @@
+module Migrate
+  module Services
+    class ChildWorkParser
+
+      def initialize(object_hash, config, output_dir)
+        @object_hash = object_hash
+        @collection_uuids = MigrationHelper.get_collection_uuids(config['collection_list'])
+
+
+        # Store parent-child relationships
+        @parent_child_mapper = Migrate::Services::IdMapper.new(File.join(output_dir, 'parent_child.csv'), 'parent', 'children')
+        # Progress tracker for objects migrated
+        @object_progress = Migrate::Services::ProgressTracker.new(File.join(output_dir, 'object_progress.log'))
+      end
+
+      def find_children
+        start_time = Time.now
+        # build parent hash
+        puts "[#{Time.now.to_s}] Building parent-child relationships"
+
+        @collection_uuids.each do |uuid|
+          metadata = Nokogiri::XML(File.open(@object_hash[uuid]))
+
+          child_works = Array.new
+
+          # RDF information
+          cdr_model_type = ''
+          rdf_version = metadata.xpath("//rdf:RDF", MigrationConstants::NS).last
+          if rdf_version
+            # Check if aggregate work
+            if rdf_version.to_s.match(/hasModel/)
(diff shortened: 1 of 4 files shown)


lib/tasks/migrate/services/child_work_parser.rb
```diff
@@ -0,0 +1,75 @@
...
+          rdf_version = metadata.xpath("//rdf:RDF", MigrationConstants::NS).last
+          if rdf_version
+            # Check if aggregate work
+            if rdf_version.to_s.match(/hasModel/)
+              cdr_model_type = rdf_version.xpath('rdf:Description/*[local-name() = "hasModel"]/@rdf:resource', MigrationConstants::NS).map(&:text)
+            end
+
+            # Create lists of attached files and children
+            if rdf_version.to_s.match(/resource/)
+              contained_objects = rdf_version.xpath("rdf:Description/*[not(local-name()='originalDeposit') and not(local-name() = 'defaultWebObject') and contains(@rdf:resource, 'uuid')]", MigrationConstants::NS)
```
bbpennel: It seems like it would be safer to do `local-name()='contains'`. Simple objects (not AggregateWorks) have a few other properties that would have uuids in their values, and we may not be able to guarantee that there aren't other properties that happen to have uuids for Aggregates.
I'm partially suggesting this, since it may be helpful to log when a migrated object thinks it contains something, but that child didn't make it into the object_hash, since that could indicate that an object is missing.

lib/tasks/migrate/services/child_work_parser.rb
```diff
@@ -0,0 +1,75 @@
...
+              contained_objects.each do |contained_file|
+                tmp_uuid = MigrationHelper.get_uuid_from_path(contained_file.to_s)
+                if (cdr_model_type.include? 'info:fedora/cdr-model:AggregateWork') && (!@object_hash[tmp_uuid].blank?) && (tmp_uuid != uuid)
```
bbpennel: `!@object_hash[tmp_uuid].blank?` might be more readable as `@object_hash.key?(tmp_uuid)` unless you are worried about there being keys with empty values.

lib/tasks/migrate/services/child_work_parser.rb
```diff
@@ -0,0 +1,75 @@
...
+                end
+              end
+
+              representative = rdf_version.xpath('rdf:Description/*[local-name() = "defaultWebObject"]/@rdf:resource', MigrationConstants::NS).to_s.split('/')[1]
+              if representative
+                representative_uuid = MigrationHelper.get_uuid_from_path(representative)
+                child_works -= [representative_uuid]
+
+                # Record that primary object was completed
+                @object_progress.add_entry(representative_uuid)
```
bbpennel: If i'm understanding correctly, the logic is that we are recording the primary object as already migrated (even though it wasn't) to prevent the child work for it from being generated? That should work.

smithjp: Yes, that was what I was trying to do here.  It does seem to skip the items that should be files during migration.

PR #317, opened 2019-03-12
Store deposit agreement acceptance in fedora

smithjp:
https://jira.lib.unc.edu/browse/HYC-531


app/forms/single_value_form.rb
```diff
@@ -90,7 +91,13 @@ def self.model_attributes(form_params)
     edtf_form_update(attrs, :date_created)
     edtf_form_update(attrs, :date_issued)
 
-    attrs
+    # Log deposit agreement acceptance
+    if agreement != '0'
+      (attrs[:deposit_agreement] ||= []) << "#{agreement} accepted the deposit agreement on #{Time.now}"
+    end
+    attrs[:deposit_agreement] = attrs[:deposit_agreement].join(',').split(',')
```
bbpennel: Hmm, i don't understand what this is doing, it seems like the result of the joining and splitting would be no change.

app/views/hyrax/base/_form_progress.html.erb
```diff
@@ -33,8 +33,13 @@
     <% if ::Flipflop.show_deposit_agreement? %>
       <% if ::Flipflop.active_deposit_agreement_acceptance? %>
         <label>
-          <%= check_box_tag 'agreement', 1, f.object.agreement_accepted, required: true %>
-          <%= t('hyrax.active_consent_to_agreement') %><br />
+          <%= f.input :agreement,
```
bbpennel: I believe this needs to be marked as a required field in order to get the text box to highlight as it did before.
Also, this is resulting in the style changes: it is no longer bolded and there is a lot of space between `I have read and agree to the` and `Deposit Agreement`. The following css changes should get it to look almost the same:
```
.form-progress .form-group {
    margin-bottom: 0;
}
.form-progress .checkbox {
    margin-bottom: 0;
    margin-top: 0;
}
.form-progress .checkbox label {
    font-weight: bold;
}
```

app/forms/single_value_form.rb
smithjp: The way the hidden field had to be set up, the previous deposit agreement statements are passed into and out of the form as a comma-separated string.  This adds the new statement to that string and then converts it back into an array to be stored in fedora.  I'll add a comment in the code to document this, too.

PR #330, opened 2019-03-15
Send deletion request emails to the admin group

smithjp:
https://jira.lib.unc.edu/browse/HYC-551


bbpennel:
I'm getting weird results from this branch on qa. As the depositor and deletion requester, I am getting both notifications (for the request and after the approval) even though i'm not an admin. Rebekah is not getting any notifications, but is able to approve the deletion.

I think I did everything necessary to make the change live, including redeploying hyrax.yml using the updates-2019-03-19 branch,  ran reload_workflows.yml and restarted sidekiq and hyc.

smithjp:
It looks like because the deleting role was added after the admin sets were created, there may not be a deleting role for any admin set.  I'll double check this and see if I can fix it.

PR #334, opened 2019-03-19
Set child admin set

bbpennel:


smithjp:
I tested this on my VM, and everything looks good.

PR #338, opened 2019-03-20
Update devise to 4.6.1 for security fixes

smithjp:


smithjp:
Also, https://gitlab.lib.unc.edu/cdr/hyc-gems/merge_requests/12

PR #340, opened 2019-03-25
merge honors and academic concentrations vocabularies

lfarrell:
https://jira.lib.unc.edu/browse/HYC-537


app/views/records/edit_fields/_academic_concentration.html.erb
```diff
@@ -1,8 +1,26 @@
 <%# app/views/records/edit_fields/_academic_concentration.html.erb %>
+<%
+  case params[:controller]
+  when 'hyrax/honors_theses'
+    value = 'honors'
```
bbpennel: Minor, but `value` to me implies the value of the field rather than which work type or vocabulary subset to use.

config/authorities/academic_concentration.yml
```diff
@@ -1,41 +1,226 @@
...
+    active: honors
+  - id: Political Science
+    term: Political Science
+    active: honors
+  - id: Psychology
+    term: Psychology
+    active: honors
   - id: Public Health
     term: Public Health
+    active: honors and masters
```
bbpennel: My feeling is we should use a different property, since `active` already has meaning and we may want to be able to disable terms separately. Maybe something like `tags` to be general, or `work_type` if we want a more specific field.

Also, it should be fine to leave out "and".

I was trying to decide if using an array would be clearer, so:
```
tags:
- honors
- masters    
```
in which case it'd be a lookup in an array instead of a regex, but that may just be a personal preference, so i'll leave that up to you.

PR #343, opened 2019-03-26
migrate date created for works and files

smithjp:
* https://jira.lib.unc.edu/browse/HYC-593
* migrate date_created for works and filesets
* display date_created as date uploaded when present
* remove date_created from all forms


lib/tasks/migrate/services/metadata_parser.rb
```diff
@@ -27,14 +27,12 @@ def parse
 
         work_attributes['contained_files'] = Array.new(0)
 
-        # get the date_uploaded
+        # get the date_created
         date_uploaded_string = metadata.xpath("//foxml:objectProperties/foxml:property[contains(@NAME, 'model#createdDate')]/@VALUE", MigrationConstants::NS).to_s
-        date_uploaded = DateTime.strptime(date_uploaded_string, '%Y-%m-%dT%H:%M:%S.%N%Z').strftime('%Y-%m-%d') unless date_uploaded_string.nil?
-        work_attributes['date_uploaded'] = (Date.try(:edtf, date_uploaded) || date_uploaded).to_s
+        work_attributes['date_created'] = DateTime.strptime(date_uploaded_string, '%Y-%m-%dT%H:%M:%S.%N%Z').strftime('%Y-%m-%d') unless date_uploaded_string.nil?
```
bbpennel: I know its been this way a while, but you can parse iso8601 dates with:
`Time.iso8601(date_uploaded_string).strftime('%Y-%m-%d')`

Maybe a bit broader question, but why are we changing the precision of how these dates are stored versus formatting them on display? `dc:dateSubmitted` is stored in fedora with `2019-03-21T12:48:51.597409138+00:00`, same with `dc:dateModified`.

smithjp: I think we had decided at one point that date_created should be one of the dates stored in edtf format, so it needed to be a date instead of a datetime.  We probably should check with others to see if we want to change that now.

lib/active_fedora/rdf/indexing_service.rb
```diff
@@ -164,8 +164,11 @@ def build_person_display(field_key, people)
             display_text << "ORCID: #{Array(person['orcid']).first}" if !Array(person['orcid']).first.blank?
             @orcid_label.push(Array(person['orcid']))
 
-            display_text << "Affiliation: #{split_affiliations(person['affiliation']).join(', ')}" if !Array(person['affiliation']).first.blank?
-            @affiliation_label.push(Array(person['affiliation']))
+            affiliations = split_affiliations(person['affiliation'])
```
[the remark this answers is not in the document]
smithjp: No, but I was seeing it as I was testing things.  I thought it had already been fixed, and I fixed it locally.  I must have accidentally included it in this pr.  I can take it out.

PR #346, opened 2019-03-28
turn off emails

smithjp:
* https://jira.lib.unc.edu/browse/HYC-415
* turn off emails and in-app notifications for the migration
* tests will fail


smithjp:
That seems like a good plan.  I'll make the changes.

PR #352, opened 2019-04-03
mime type check

smithjp:
* https://jira.lib.unc.edu/browse/HYC-604

### DIFF
--- a/app/controllers/concerns/hydra/controller/download_behavior.rb
+++ b/app/controllers/concerns/hydra/controller/download_behavior.rb
@@ -84,7 +84,8 @@ module Hydra
       def file_name
         filename = params[:filename] || file.original_name || (asset.respond_to?(:label) && asset.label) || file.id
         extension = MimeTypeService.label(file.mime_type)
-        if !filename.match?(/.*#{extension}\z/)
+        existing_extension = filename.split('.').last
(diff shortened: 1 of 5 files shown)


app/controllers/concerns/hydra/controller/download_behavior.rb
```diff
@@ -84,7 +84,8 @@ def content_options
       def file_name
         filename = params[:filename] || file.original_name || (asset.respond_to?(:label) && asset.label) || file.id
         extension = MimeTypeService.label(file.mime_type)
-        if !filename.match?(/.*#{extension}\z/)
+        existing_extension = filename.split('.').last
```
bbpennel: Here is a suggested refactoring:
```
file_parts = filename.split('.')
existing_extension = file_parts.length > 1 ? file_paths.last : nil
if MimeTypeService.valid?(existing_extension).blank?
  extension = MimeTypeService.label(file.mime_type)
  "#{filename}.#{extension}"
else
  filename
end
```
One case that could cause problems in the existing implementation would be if the filename happened to be a valid extension too. So if the filename were "ttl" then the returned value would be "ttl" rather than "ttl.txt".

I also don't think we need to populate `extension` unless the existing extension is invalid. If the filename was "file.txt", then the validity check would cause it to be returned by itself.

bbpennel:
This looks good over all, had one suggestion regarding the conditional for selecting when to add an extension.

PR #353, opened 2019-04-04
Date information in citations is incorrect

lfarrell:
https://jira.lib.unc.edu/browse/HYC-619


app/helpers/hyrax/citations_behaviors/publication_behavior.rb
```diff
@@ -1,15 +1,25 @@
...
 module Hyrax
   module CitationsBehaviors
     module PublicationBehavior
       include Hyrax::CitationsBehaviors::CommonBehavior
       def setup_pub_date(work)
-        first_date = work.date_created.first if work.date_created
+        if work.respond_to?(:date_issued)
+          file_date = work.date_issued
+          first_date = file_date.respond_to?(:first) ? file_date.first : file_date
+        elsif work.respond_to?(:date_created)
```
bbpennel: Is it possible that something would respond to date_issued but not have a value? In which case we might want to fall back to date_created?

lfarrell: I guess not. I thought at least one of the forms didn't use it, but that doesn't seem to be the case.

bbpennel:
This looks good, but there are still some failing tests.

I'm noticing that the MLA citations look goofy when there is no publisher, but perhaps that's a different ticket. For example, `In the Works: Conflict In the Context of Everyday Practice. , 1986. `

PR #354, opened 2019-04-09
Authors don't display in citations

lfarrell:
https://jira.lib.unc.edu/browse/HYC-621


app/helpers/hyrax/citations_behaviors/name_behavior.rb
```diff
@@ -0,0 +1,57 @@
+module Hyrax
```
smithjp: Was this file added from the gem?

lfarrell: Yes

lfarrell: Need to add the override header :)

PR #358, opened 2019-04-16
Don't have admins check agreement box on edit

lfarrell:
Form won't work if something isn't submitted, so just hide it.
https://jira.lib.unc.edu/browse/HYC-624


smithjp:
This prevents admins from adding files on edit, which could be fine.

PR #370, opened 2019-04-25
Uv viewer settings

lfarrell:
Hide metadata panel. Same metadata shows below the image anyway.
Hide settings button
Remove download option from viewer. Add download link below viewer.

https://jira.lib.unc.edu/browse/HYC-647
https://jira.lib.unc.edu/browse/HYC-640

### DIFF
--- a/app/views/hyrax/base/show.html.erb
+++ b/app/views/hyrax/base/show.html.erb
@@ -29,6 +29,15 @@
           <% end %>
           <div class="col-sm-3 text-center">
             <%= render 'representative_media', presenter: @presenter, viewer: false unless @presenter.iiif_viewer? %>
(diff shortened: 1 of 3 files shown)


app/views/hyrax/base/show.html.erb
```diff
@@ -29,6 +29,15 @@
           <% end %>
           <div class="col-sm-3 text-center">
             <%= render 'representative_media', presenter: @presenter, viewer: false unless @presenter.iiif_viewer? %>
```
bbpennel: What would happen if you removed the new block added in this PR and changed this line to:
`<%= render 'representative_media', presenter: @presenter, viewer: false %>`
so that it would show the representative media section regardless of the viewer?

PR #371, opened 2019-04-25
Hyc 618 missing thumbnails

smithjp:
* https://jira.lib.unc.edu/browse/HYC-618

### DIFF
--- a/lib/mini_magick/image/info.rb
+++ b/lib/mini_magick/image/info.rb
@@ -1,3 +1,4 @@
+# [hyc-override] Remove warnings before parsing image data
 require "json"
 
 module MiniMagick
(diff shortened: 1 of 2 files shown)


lib/mini_magick/image/info.rb
```diff
@@ -40,9 +41,14 @@ def clear
         @info.clear
       end
 
+      # [hyc-override] remove warnings and print errors
       def cheap_info(value)
         @info.fetch(value) do
-          format, width, height, size = self["%m %w %h %b"].split(" ")
+          if !self["%m %w %h %b"].blank? && self["%m %w %h %b"].include?('Warning')
```
bbpennel: I'm not sure how expensive `self["%m %w %h %b"]` is (I'm aware this method is called `cheap_info`, but it seems like the call goes a bit deep into the identifier), but it might be prudent to capture the response from it once and reuse that value. It'll also be a little easier to maintain and read as well.

PR #374, opened 2019-04-29
Fix homepage margins on large screens

lfarrell:
homepage


bbpennel:
The sections of the page should be the full width rather, so you will likely need to switch to padding instead of margins. See:
https://jira.lib.unc.edu/browse/HYC-668

PR #381, opened 2019-05-09
Fix allied affiliations 

bbpennel:


lib/tasks/remediate_person_affiliations.rake
```diff
@@ -0,0 +1,64 @@
...
+              result = p.affiliation.delete?('College of Arts and Sciences')
+              # Use this for testing if need to add in a second affiliation
+              # puts "ADDING"
+              # p.affiliation << 'College of Arts and Sciences'
+              # result = true
+            else
+              result = nil
+            end
+            
+            p_hash = JSON.parse(p.to_json)
```
lfarrell: Is it necessary to add the deleted metadata to an array? It doesn't seem to get used for anything.

PR #386, opened 2019-05-15
Remediation tasks

bbpennel:


lib/tasks/change_admin_set.rake
```diff
@@ -0,0 +1,25 @@
...
+  raise "Not a valid admin set ID: #{admin_set_id}" if admin_set.nil?
+  
+  puts "Switching works to use admin set #{admin_set.title.first} (#{admin_set.id})"
+
+  lines = File.readlines(id_list_path)
+  total = lines.length
+  lines.each_with_index do |line, index|
+    start = Time.now
+    
+    model, id = line.split(',').collect(&:strip)
```
lfarrell: What does model get used for?

lib/tasks/inherit_access_and_change_admin_set.rake
```diff
@@ -0,0 +1,100 @@
...
+  raise "Not a valid admin set ID: #{admin_set_id}" if admin_set.nil?
+  
+  lines = File.readlines(id_list_path)
+  total = lines.length
+  puts "Updating the access restrictions and admin set of the children of #{total} works"
+  
+  lines.each_with_index do |line, index|
+    start = Time.now
+    
+    model, id = line.split(',').collect(&:strip)
```
lfarrell: Same here, not sure what model is used for.

lib/tasks/update_visibility.rake
```diff
@@ -0,0 +1,39 @@
...
+  start_total = Time.now
+  
+  lines = File.readlines(id_list_path)
+  total = lines.length
+  puts "Updating access restrictions for #{total} works to #{new_visibility}"
+  
+  lines.each_with_index do |line, index|
+    start = Time.now
+    
+    model, id = line.split(',').collect(&:strip)
```
lfarrell: Same here with model variable

lib/tasks/change_admin_set.rake
bbpennel: Nothing apparently :) It got copy and pasted from another task I was basing things off of. Apparently you don't need that info for lookups this way. I'll remove it

PR #389, opened 2019-05-17
HYC-722 - Add identifier linking to hyc record in OAI

bbpennel:
https://jira.lib.unc.edu/browse/HYC-722

### DIFF
--- a/app/models/concerns/blacklight/document/dublin_core.rb
+++ b/app/models/concerns/blacklight/document/dublin_core.rb
@@ -55,7 +55,13 @@ module Blacklight::Document::DublinCore
           end
         end
       end
+      # Add an identifier linking to the record page in hyc if no DOI is present
+      if doi.blank?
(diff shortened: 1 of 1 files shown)


app/models/concerns/blacklight/document/dublin_core.rb
```diff
@@ -55,13 +55,23 @@ def export_as_oai_dc_xml
...
+      end
     end
+
     xml.target!
   end
 
   alias_method :export_as_xml, :export_as_oai_dc_xml
   alias_method :export_as_dc_xml, :export_as_oai_dc_xml
 
+  def deleted?
```
lfarrell: Does this get used? I'm not seeing it anywhere

bbpennel: The usage isn't in our codebase, its over here:
https://github.com/code4lib/ruby-oai/blob/52aa1cc0d3d2a250f0d4480e0be32f13d8c97692/lib/oai/provider/response/record_response.rb#L77
Having this method present causes the `status="deleted"` header to be added.

PR #409, opened 2019-06-17
proquest ingest

smithjp:
* https://jira.lib.unc.edu/browse/HYC-766

### DIFF
--- a/lib/tasks/proquest_ingest.rake
+++ b/lib/tasks/proquest_ingest.rake
@@ -36,41 +36,55 @@ namespace :proquest do
       puts "Unpacking #{package}"
       @file_last_modified = ''
       unzipped_package_dir = extract_proquest_files(package)
-      metadata_files = Dir.glob("#{unzipped_package_dir}/**/*")
 
-      metadata_files.sort.each do |file|
-        if File.file?(file)
-          if file.match('.xml')
-            metadata_fields = proquest_metadata(file, metadata_dir)
+      # get all files in unzipped directory (should be 1 pdf and 1 xml)
+      metadata_file = Dir.glob("#{unzipped_package_dir}/*_DATA.xml")
+      if metadata_file.count == 1
+        metadata_file = metadata_file.first.to_s
+      else
+        puts "#{unzipped_package_dir} has more than 1 xml file"
+        next
+      end
+      pdf_file = Dir.glob("#{unzipped_package_dir}/*.pdf")
+      if pdf_file.count == 1
+        pdf_file = pdf_file.first.to_s
+      else
+        puts "#{unzipped_package_dir} has more than 1 pdf file"
+        next
+      end
 
-            puts "Number of files: #{metadata_fields[:files].count.to_s}"
+      if File.file?(metadata_file)
+        # only use xml file for metadata extraction
+        metadata_fields = proquest_metadata(metadata_file)
 
-            # create deposit record
-            deposit_record = DepositRecord.new(@deposit_record_hash)
-            deposit_record[:manifest] = nil
-            deposit_record[:premis] = nil
-            deposit_record.save!
+        puts "Number of files: #{metadata_fields[:files].count.to_s}"
 
-            resource = proquest_record(metadata_fields[:resource])
-            resource[:deposit_record] = deposit_record.id
-            resource.save!
+        # create deposit record
+        deposit_record = DepositRecord.new(@deposit_record_hash)
+        deposit_record[:manifest] = nil
+        deposit_record[:premis] = nil
+        deposit_record.save!
 
-            ingest_proquest_files(resource: resource,
-                         files: metadata_fields[:files],
-                         metadata: metadata_fields[:resource],
-                         zip_dir_files: metadata_files)
+        # create disseration record
+        resource = proquest_record(metadata_fields[:resource])
+        resource[:deposit_record] = deposit_record.id
+        resource.save!
 
-            # Attach metadata file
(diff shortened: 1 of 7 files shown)


lib/tasks/proquest_ingest.rake
```diff
@@ -6,43 +6,28 @@ namespace :proquest do
...
+                             deposit_package_type: config['deposit_type'],
+                             deposit_package_subtype: config['deposit_subtype'],
+                             deposited_by: @depositor_onyen }
 
-    metadata_dir = args[:directory]
-    migrate_proquest_packages(metadata_dir)
+    migrate_proquest_packages(config['package_dir'])
   end
 
   def migrate_proquest_packages(metadata_dir)
```
bbpennel: It appears that `@temp` is recursively deleted after each package, and then recreated as a side effect of unzipping the package in `extract_proquest_files`. That worries me a little, it'd be nice not to risk deleting extra stuff (like if I configured it to use `/tmp`).

Hopefully I'm not misunderstanding how things are working, but things might be clearer/safer if the body of the loop at line 35 more explicitly dealt with a single package. Here's an example of what I was thinking:
```
    proquest_packages.each do |package|
      puts "Unpacking #{package}"
      @file_last_modified = ''
      unzipped_package_dir = extract_proquest_files(package)
      # We're only interested in what's happening in the current package
      metadata_files = Dir.glob("#{unzipped_package_dir}/*_DATA.xml")
      # there should only be one metadata file at the base of the package
      metadata_file = metadata_files[0]

      metadata_fields = proquest_metadata(file, metadata_dir)
      # ... create objects and do metadata stuff like currently
      # ...

      # explicitly deleting the unzipped package
      FileUtils.rm_rf(unzipped_package_dir)
    end
```

`extract_proquest_files` would need to return the name of the directory it creates, or it could return the path to the metadata file and you could skip the globbing.

bbpennel: Sorry to review things outside of the current PR, but this is regarding line 44. `proquest_metadata` does not currently use metadata_dir, so it can likely be dropped from the signature.

lib/tasks/proquest_ingest.rake
```diff
@@ -60,13 +45,31 @@ namespace :proquest do
...
             resource.save!
 
             ingest_proquest_files(resource: resource,
                          files: metadata_fields[:files],
                          metadata: metadata_fields[:resource],
                          zip_dir_files: metadata_files)
+
+            # Attach metadata file
+            fileset_attrs = { 'title' => [File.basename(file)],
+                              'visibility' => Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PRIVATE }
```
bbpennel: Wonder if we can drop this `visibility` setting since it sounds like it wasn't working, and its happening again below?

lib/tasks/proquest_ingest.rake
```diff
@@ -275,7 +279,55 @@ namespace :proquest do
...
+    end
+
+    person_hash
+  end
+
   def format_name(person)
-    (person.xpath('DISS_surname').text+', '+person.xpath('DISS_fname').text+' '+person.xpath('DISS_middle').text).split.join(' ')
+    name_parts = []
+    name_parts << person.xpath('DISS_surname').text
+    name_parts << (person.xpath('DISS_fname').text+' '+person.xpath('DISS_middle').text).split.join(' ')
```
bbpennel: Is the split and join to trim off the trailing space if there's no middle name? Would `.strip` work instead?

lib/tasks/proquest_ingest.rake
```diff
@@ -106,14 +109,14 @@ namespace :proquest do
...
-    if resource['embargo_release_date'].blank?
+    fileset_metadata = file_record(resource)
+
+    if fileset_metadata['embargo_release_date'].blank?
       fileset_metadata.except!('embargo_release_date', 'visibility_during_embargo', 'visibility_after_embargo')
     end
     file_set = FileSet.create(fileset_metadata)
-    actor = Hyrax::Actors::FileSetActor.new(file_set, User.find_by_email(@depositor_email))
-    actor.create_metadata(resource)
+    actor = Hyrax::Actors::FileSetActor.new(file_set, User.where(uid: @depositor_onyen).first)
```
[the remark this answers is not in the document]
smithjp: It looks like it is only using the user object once in the script and the onyen everywhere else.  I'll make sure that is the case and switch to passing the user object if it is being used more than once.

lib/tasks/proquest_ingest.rake
```diff
@@ -6,52 +6,37 @@ namespace :proquest do
...
 
   def migrate_proquest_packages(metadata_dir)
     proquest_packages = Dir.glob("#{metadata_dir}/*.zip")
     proquest_packages.each do |package|
       puts "Unpacking #{package}"
       @file_last_modified = ''
-      extract_proquest_files(package)
-      metadata_files = Dir.glob("#{@temp}/**/*")
+      unzipped_package_dir = extract_proquest_files(package)
+      metadata_files = Dir.glob("#{unzipped_package_dir}/**/*")
```
bbpennel: my recollection is that in these packages there is only ever going to be one DATA.xml file and it will be in the top level of the package. So you might be able to save some steps by doing `Dir.glob("#{unzipped_package_dir}/*_DATA.xml")` and then you could likely drop lines 41-43 since you wouldn't need to iterate, it'd only give you .xml files, and i don't think there'd ever be a directory matching this pattern in this controlled proquest environment.

PR #411, opened 2019-06-20
ga stats redirect

smithjp:
* https://jira.lib.unc.edu/browse/HYC-762


smithjp:
I forgot to add override comments.  If we decide to ever merge this, I'll start from a fresh branch and commit the originals and the changes.

app/models/hyrax/download.rb
```diff
@@ -0,0 +1,9 @@
+module Hyrax
+  class Download
+    extend ::Legato::Model
+
+    metrics :totalEvents
+    dimensions :eventCategory, :eventAction, :eventLabel, :date
+    filter :for_file, &->(id) { contains(:eventLabel, id) }
```
smithjp: changed 'matches' to 'contains'

app/models/hyrax/file_download_stat.rb
```diff
@@ -0,0 +1,42 @@
...
+        return []
+      end
+      profile.hyrax__download(sort: 'date',
+                              start_date: start_date,
+                              end_date: Date.yesterday)
+          .for_file(file.id)
+    end
+
+    # this is called by the parent class
+    # [hyc-override] include previous GA data
```
smithjp: added lines 23-37

app/models/hyrax/statistic.rb
```diff
@@ -0,0 +1,82 @@
...
+      # see Legato::ProfileMethods.method_name_from_klass
+      # [hyc-override] include previous GA data
+      def ga_statistics(start_date, object)
+        path = polymorphic_path(object)
+        profile = Hyrax::Analytics.profile
+        unless profile
+          Rails.logger.error("Google Analytics profile has not been established. Unable to fetch statistics.")
+          return []
+        end
+
```
smithjp: added lines 38-50 and modified 52

app/models/hyrax/file_download_stat.rb
```diff
@@ -0,0 +1,42 @@
...
+    def filter(file)
+      filter_id = file.id
+
+      # check if file was migrated
+      if ENV.has_key?('REDIRECT_FILE_PATH') && File.exist?(ENV['REDIRECT_FILE_PATH'])
```
bbpennel: if i’m understanding correctly it is looking up the old id and querying for both? That seems reasonable.

The redirect look up logic is probably repeated in 3 places now though (two in this PR, one in the actual redirect script). Would it be possible to consolidate these lookups into a single method? And is this happening on every page load, potentially multiple times? Wondering if it would make sense to use some kind of singleton cache to hold the mapping rather than reading the 3mb csv file so often? Although I'm sure the file gets well cached.

PR #413, opened 2019-06-25
Update index for optional people fields

lfarrell:
https://jira.lib.unc.edu/browse/HYC-782

### DIFF
--- a/app/assets/javascripts/person_objects.js
+++ b/app/assets/javascripts/person_objects.js
@@ -99,14 +99,7 @@ $(document).on('turbolinks:load', function () {
     
     function updateCloningRow($cloning_row, current_index) {
         var regex = /\d+/;
-        var next_index;
-
-        // Required people seem to have their index updated on page load, Optional people don't, hence the double increment
-        if ($cloning_row.find("[id$='name']").hasClass('optional')) {
-            next_index = current_index + 2;
-        } else {
-            next_index = current_index + 1;
-        }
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/person_objects.js
```diff
@@ -99,7 +99,14 @@ $(document).on('turbolinks:load', function () {
     
     function updateCloningRow($cloning_row, current_index) {
         var regex = /\d+/;
-        var next_index = current_index + 1;
+        var next_index;
+
+        // Required people seem to have their index updated on page load, Optional people don't, hence the double increment
+        if ($cloning_row.find("[id$='name']").hasClass('optional')) {
```
bbpennel: This is interesting, any idea what is causing this weird index setting?

lfarrell: Maybe because required fields are visible on page load, but optional fields aren't. Though that's just a working hypothesis.

app/assets/javascripts/person_objects.js
```diff
@@ -121,7 +121,7 @@ $(document).on('turbolinks:load', function () {
     }
 
     function getPersonRows(selector) {
-        return $('#' + selector + ' div.' + selector).not(':hidden');
+        return $('#' + selector + ' div.' + selector).not('.cloning, .cloning div');
```
bbpennel: Good work!

PR #416, opened 2019-06-28
Add group permissions to ingest rake tasks

smithjp:
* https://jira.lib.unc.edu/browse/HYC-772


lib/tasks/migrate/services/ingest_service.rb
```diff
@@ -220,12 +228,17 @@ def ingest_records
       end
 
       def create_fileset(parent: nil, resource: nil, file: nil)
+        permissions_attributes = resource['permissions_attributes']
+        resource.delete('permissions_attributes')
```
bbpennel: This PR is good, my only thought is that adding "permissions_attributes" and then deleting it is a little confusing, wonder if passing it as another method param would have been clearer?

PR #426, opened 2019-07-19
Attach artwork button

lfarrell:
https://jira.lib.unc.edu/browse/HYC-807

### DIFF
--- a/app/views/hyrax/base/_show_actions.html.erb
+++ b/app/views/hyrax/base/_show_actions.html.erb
@@ -32,7 +32,7 @@
     <% if presenter.valid_child_concerns.length > 0 %>
       <div class="btn-group">
 
-        <% if current_user&.admin? || (@presenter.workflow.present? && !@presenter.workflow.is_mfa?) %>
+        <% if current_user&.admin? || !@presenter.workflow.is_mfa? %>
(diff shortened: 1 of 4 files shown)


app/views/hyrax/base/_show_actions.html.erb
```diff
@@ -53,9 +52,11 @@
               <% end %>
             </ul>
         <% end %>
+        <% if (!current_user&.admin? && @presenter.workflow.present? && @presenter.workflow.is_mfa_in_review?) || (@presenter.workflow.is_mfa? && current_user&.admin_unit_manager?('art_mfa_manager'))  %>
```
bbpennel: Hmm, `!current_user&.admin?` would let `current_user = nil` through if the MFA was in review. Can you actually get to this partial with a nil user?

If you can't get to this partial with a nil user, you could lose a few `&`'s. You could also reduce this conditional a bit and make it clearer that you can't have "attach child" and "attach artwork" at the same time by using an `else if`:
```
elsif @presenter.workflow.present? && (@presenter.workflow.is_mfa_in_review? || current_user.admin_unit_manager?('art_mfa_manager'))
```
since the else implies that the user isn't an admin and there's either no workflow or the workflow is an mfa one.

lfarrell: Shortened version

PR #430, opened 2019-07-26
Proquest ingest update

smithjp:
* Updates department and degree mappings for proquest records
  * Some departments were not in existing authority
  * One department was mapped to the wrong key
  * It looks like the proquest abbreviation for the doctor of education degree changed from E.D.D. to D.E., so we now have both in the mapping
  * Will store non-mapping departments in fedora in the future for easier remediation
* Adds log statements to help identify new records and log zip file errors
* Fixes artwork feature test

### DIFF
--- a/config/authorities/proquest_departments.yml
+++ b/config/authorities/proquest_departments.yml
@@ -75,7 +75,7 @@ terms:
   - id: 'Dentistry: Endodontics'
     term: Department of Endodontics
   - id: Ecology
-    term: Curriculum for the Environment and Ecology
+    term: Curriculum in Environment and Ecology
   - id: Economics
     term: Department of Economics
   - id: Education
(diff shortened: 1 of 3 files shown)


config/authorities/proquest_departments.yml
```diff
@@ -242,3 +242,13 @@ terms:
     term: Department of Statistics and Operations Research
   - id: Toxicology
     term: Curriculum in Toxicology
+  - id: Clinical Rehabilitation and Mental Health Counseling
+    term: Division of Clinical Rehabilitation and Mental Health Counseling
+  - id: English and Comparative Literature
+    term: Department of English and Comparative Literature
+  - id: Education (Educational Leadership )
```
bbpennel: Do they really have a trailing space before the `)`?

smithjp: they really do

PR #433, opened 2019-07-30
Even more proquest ingest updates

smithjp:
* updating fileset title
* adding supplemental files (whoops)

### DIFF
--- a/lib/tasks/proquest_ingest.rake
+++ b/lib/tasks/proquest_ingest.rake
@@ -102,9 +102,10 @@ namespace :proquest do
           file_path = unzipped_file_list.find { |e| e.match(f.to_s) }
           if file_path.blank?
             puts "[#{Time.now}][#{id}] cannot find #{f.to_s}"
+            next
           end
 
-          if !file_path.nil? && File.file?(file_path)
+          if File.file?(file_path)
             file_set = ingest_proquest_file(parent: resource,
                                             resource: metadata_fields[:resource].merge({title: [f]}),
                                             f: file_path)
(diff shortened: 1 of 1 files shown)


lib/tasks/proquest_ingest.rake
```diff
@@ -118,7 +135,7 @@ namespace :proquest do
       Zip::File.open(file) do |zip_file|
         zip_file.each do |f|
           if f.name.match(/DATA.xml/)
-            @file_last_modified = Date.strptime(zip_file.get_entry(f).as_json['time'].split('T')[0],"%Y-%m-%d")
+            @file_last_modified = Date.strptime(zip_file.get_entry(f).as_json['time'].split('T')[0],"%Y-%m-%d").year
```
bbpennel: Not really that important, but was there a functional reason getting the year out needed to move from the `graduation_year` assignment to here? Seems like it might make sense to keep the full last modified date or go straight to graduation_year.

lib/tasks/proquest_ingest.rake
```diff
@@ -84,17 +84,34 @@ namespace :proquest do
...
+        # get list of all files in unzipped proquest package
+        unzipped_file_list = Dir.glob("#{unzipped_package_dir}/**/*.*")
+
+        ordered_members = []
+        metadata_fields[:files].each do |f|
+          puts "[#{Time.now}] trying...#{f.to_s}"
+
+          file_path = unzipped_file_list.find { |e| e.match(f.to_s) }
+          if file_path.blank?
+            puts "[#{Time.now}][#{id}] cannot find #{f.to_s}"
```
bbpennel: Might as well add a `break` in here, don't think you would want to proceed with a blank file path. You could then also drop the `!file_path.nil?` part from the next `if`

lib/tasks/proquest_ingest.rake
smithjp: No, I had changed it to be in another place, and then moved it back incorrectly.  I'll fix it.

PR #436, opened 2019-08-05
HYC-773 - People from Fedora and more logging

bbpennel:


app/helpers/hyc/doi_create.rb
```diff
@@ -124,34 +126,42 @@ def format_data(record)
     end
 
     def create_doi(record)
-      Rails.logger.info "Creating DOI for #{record['id']}"
-      response = doi_request(format_data(record))
+      puts "Creating DOI for #{record['id']}"
+      record_data = format_data(record)
+      puts record_data.inspect
```
smithjp: do we want this to be logged?

bbpennel: nope :)

PR #454, opened 2019-09-17
Redirect dois to correct pages

smithjp:
* https://jira.lib.unc.edu/browse/HYC-862

### DIFF
--- a/spec/features/redirect_doi_url_spec.rb
+++ b/spec/features/redirect_doi_url_spec.rb
@@ -1,0 +1,30 @@
+require 'rails_helper'
+include Warden::Test::Helpers
+
+RSpec.feature 'redirect doi urls', js: false do
+  context 'for models with two-word names' do
+    let(:honors_thesis) { HonorsThesis.create(title: ['honors thesis doi test'], visibility: 'open')}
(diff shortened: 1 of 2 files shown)


spec/features/redirect_doi_url_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+include Warden::Test::Helpers
+
+RSpec.feature 'redirect doi urls', js: false do
+  context 'for models with two-word names' do
+    let(:honors_thesis) { HonorsThesis.create(title: ['honors thesis doi test'], visibility: 'open')}
+    let(:masters_paper) { MastersPaper.create(title: ['masters paper doi test'], visibility: 'open')}
+    let(:data_set) { DataSet.create(title: ['data set doi test'], visibility: 'open')}
+    let(:scholarly_work) { ScholarlyWork.create(title: ['scholaraly work doi test'], visibility: 'open')}
+
+    scenario 'when doi links are wrong' do
```
bbpennel: Is the case of a correct URL request for these work types covered somewhere else?

smithjp: The correct urls are resolved before these redirects, which is why I put them at the bottom.  I believe the work type routes are added here: https://github.com/UNC-Libraries/hy-c/blob/master/config/routes.rb#L60

PR #457, opened 2019-10-01
Update method for determining google analytics start date

smithjp:
* https://jira.lib.unc.edu/browse/HYC-872

### DIFF
--- a/spec/presenters/hyrax/stats_usage_presenter_spec.rb
+++ b/spec/presenters/hyrax/stats_usage_presenter_spec.rb
@@ -1,0 +1,28 @@
+require 'rails_helper'
+
+RSpec.describe Hyrax::StatsUsagePresenter do
+
+  describe '#created' do
+    let(:subject) {described_class.new}
+
+    context 'for a migrated work' do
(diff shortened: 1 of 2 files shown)


spec/presenters/hyrax/stats_usage_presenter_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  describe '#created' do
+    let(:subject) {described_class.new}
+
+    context 'for a migrated work' do
+      article = Article.create(title: ['new article'], date_created: DateTime.yesterday)
+      it 'sets google analytics query start date to original date created' do
+        allow(subject).to receive(:model).and_return(article)
+
+        expect(article.date_created).not_to eq article.create_date
```
bbpennel: Are the values of date_created and create_date things that needs to be tested here? They seem like setup rather than outcome. I am wondering if this and line 23 can be removed for clarity since it seems like we are mainly wanting to verify that the stats_usage_presenter is functioning properly.

smithjp: yes, I think those lines can be removed.  they were mostly for checking the values while I was writing the tests.

PR #458, opened 2019-10-02
Hyc 847 controller tests

smithjp:
* https://jira.lib.unc.edu/browse/HYC-847


spec/controllers/hyrax/multimeds_controller_spec.rb
```diff
@@ -0,0 +1,158 @@
...
+    User.find_by_user_key('admin')
+  end
+
+  let(:admin_set) do
+    AdminSet.new(title: ['multimed admin set'],
+                 description: ['some description'],
+                 edit_users: [user.user_key])
+  end
+
+  describe '#create' do
```
bbpennel: These work type controller test methods are all very similar across types, do you think it'd make sense to use shared specs? https://relishapp.com/rspec/rspec-core/docs/example-groups/shared-examples
You could pass in the work type class and a few other details. Might make it easier to read and maintain.

spec/controllers/omniauth_controller_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+        cached_database_auth = ENV['DATABASE_AUTH']
+        cached_sso_logout_url = ENV['SSO_LOGOUT_URL']
+        ENV['DATABASE_AUTH'] = 'false'
+        ENV['SSO_LOGOUT_URL'] = 'https://shibboleth.example.com/idp/logout.jsp'
+        allow(Rails).to receive(:env).and_return(ActiveSupport::StringInquirer.new("production"))
+
+        get destroy_user_session_path
+        expect(response).to redirect_to 'https://shibboleth.example.com/idp/logout.jsp'
+
+        ENV['DATABASE_AUTH'] = cached_database_auth
```
bbpennel: Maybe move all the setup into a before and the cleanup into an after, to make sure it gets called if the test fails

spec/features/boxc_redirect_spec.rb
```diff
@@ -0,0 +1,84 @@
+require 'rails_helper'
+
+# before actions
+RSpec.feature 'boxc redirects' do
+  cached_redirect_file_path = ENV['REDIRECT_FILE_PATH']
+  cached_redirect_old_domain = ENV['REDIRECT_OLD_DOMAIN']
+  cached_redirect_new_domain = ENV['REDIRECT_NEW_DOMAIN']
+
+  before(:each) do
```
bbpennel: In this case I think you can do `before(:all)` and `after(:all)` since none of these values are changing between specs.

spec/features/boxc_redirect_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+    ENV['REDIRECT_FILE_PATH'] = cached_redirect_file_path
+    ENV['REDIRECT_OLD_DOMAIN'] = cached_redirect_old_domain
+    ENV['REDIRECT_NEW_DOMAIN'] = cached_redirect_new_domain
+  end
+
+  describe '#check_redirect' do
+    scenario 'non-boxc url' do
+      article = Article.create(title: ['test article'], visibility: 'open')
+
+      visit "#{ENV['HYRAX_HOST']}/concern/#{Array.wrap(article.class.to_s).first.downcase}s/#{article.id}"
```
bbpennel: maybe just hardcode `/concern/articles/` for readability since this isn't the totally correct way of formatting the url anyways?

spec/features/boxc_redirect_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+# before actions
+RSpec.feature 'boxc redirects' do
+  cached_redirect_file_path = ENV['REDIRECT_FILE_PATH']
+  cached_redirect_old_domain = ENV['REDIRECT_OLD_DOMAIN']
+  cached_redirect_new_domain = ENV['REDIRECT_NEW_DOMAIN']
+
+  before(:each) do
+    ENV['REDIRECT_FILE_PATH'] = 'spec/fixtures/redirect_uuids.csv'
+    ENV['REDIRECT_OLD_DOMAIN'] = 'localhost:4040/regex'
+    ENV['REDIRECT_NEW_DOMAIN'] = 'dcr-test.lib.unc.edu'
```
bbpennel: Since redirecting to outside domains appears to end up on `/`, would it be possible to use `stub_request` to fake the dcr end points you're redirecting to? We already seems to have `webmock` and are using it in other tests.

spec/features/boxc_redirect_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+    end
+
+    scenario 'boxc listContent url' do
+      visit "#{ENV['HYRAX_HOST']}/listContent/uuid:02fc897a-12b6-4b81-91e4-b5e29cb683a6?anywhere=term"
+      expect(current_path).to eq '/concern/404'
+    end
+
+    scenario 'boxc record not in hyc' do
+      # redirected to dcr-test, then to hy-c home page since testing env cannot check outside links
+      # regex for sending people to dcr expects a domain+'.', so '/regex.' was added to gsub localhost:4040
```
bbpennel: yeah this is awkward. Maybe the redirect code shouldn't expect a `.` and should instead be checking for `http(s)://domain` to make sure its replacing the right stuff

spec/features/boxc_redirect_spec.rb
```diff
@@ -0,0 +1,84 @@
+require 'rails_helper'
+
+# before actions
+RSpec.feature 'boxc redirects' do
+  cached_redirect_file_path = ENV['REDIRECT_FILE_PATH']
+  cached_redirect_old_domain = ENV['REDIRECT_OLD_DOMAIN']
+  cached_redirect_new_domain = ENV['REDIRECT_NEW_DOMAIN']
+
+  before(:each) do
+    ENV['REDIRECT_FILE_PATH'] = 'spec/fixtures/redirect_uuids.csv'
```
bbpennel: Does this need to be a relative path? It might make more sense to use a TempFile rather than a fixture file that keeps having its contents rewritten. Seems like we could end up with leftover changes to the file if a test fails.

app/controllers/application_controller.rb
```diff
@@ -13,73 +13,6 @@ class ApplicationController < ActionController::Base
   before_action :check_read_only, only: [:new, :create, :edit, :update, :destroy]
   before_action :check_redirect
 
-  # Redirect all deposit and edit requests with warning message when in read only mode
-  def check_read_only
```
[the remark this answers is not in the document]
smithjp: I was trying to figure out what to do with these since they are not controller actions.  I think moving them to a helper probably does not solve any of the access issues I was thinking about, so I am going to move them back and make them private methods since they are not public actions and should only be called from the controller classes.

app/controllers/hyrax/batch_edits_controller.rb
```diff
@@ -21,6 +21,31 @@ def edit
       @form = form_class.new(work, current_user, batch)
     end
 
+    def destroy_collection
```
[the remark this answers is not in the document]
smithjp: I made some of the methods private since that is how I am used to handling non-action methods in controllers.  The before and after hooks and any methods called from a controller action should probably be private since they would not be called outside of this class.

spec/controllers/hyrax/art_works_controller_spec.rb
```diff
@@ -2,8 +2,157 @@
...
+                 description: ['some description'],
+                 edit_users: [user.user_key])
+  end
+
+  describe '#create' do
+    let(:actor) { double(create: true) }
+
+    context 'with existing admin set' do
+      it 'is successful' do
+        artwork = Artwork.create(title: ['new artwork to be created'])
```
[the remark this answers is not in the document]
smithjp: I will move the artwork instantiation and other similar statements to `let` blocks and move the user sign_in to `before` blocks.

spec/controllers/hyrax/articles_controller_spec.rb
```diff
@@ -0,0 +1,158 @@
...
+        admin_set.save!
+        article = Article.create(title: ['work to be updated'])
+        sign_in admin_user # bypass need for permission template
+
+        get :edit, params: { id: article.id }
+        expect(response).to be_success
+      end
+    end
+
+    context 'without existing admin set' do
```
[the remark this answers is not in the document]
smithjp: We have methods on the work type controllers to check if an admin set exists before a user is allowed to create a work.  Since the servers all have admin sets, the only time this could maybe happen is on the vm, but since the method exists, I thought I should test it.

I can add tests for non-logged-in users and users without edit access on an admin set.

spec/controllers/hyrax/multimeds_controller_spec.rb
smithjp: I can move these to shared examples.  I agree that it would be easier to maintain that way.

spec/controllers/hyrax/restricted_work_type_controller_spec.rb
```diff
@@ -0,0 +1,295 @@
...
+                 description: ['some description'],
+                 edit_users: [user.user_key])
+  end
+
+  describe '#create' do
+    let(:actor) { double(create: true) }
+    let(:work) { model.create(title: ['new work to be created']) }
+
+    before(:each) do
+      @work_count = model.all.count # needs to be set before comparison
```
bbpennel: did this not work as `let(:work_count) { model.all.count }`? Just checking since its generally a bit safer not to use a global/instance variables.

spec/controllers/default_admin_sets_controller_spec.rb
```diff
@@ -210,36 +213,36 @@
   end
 
   describe "DELETE #destroy" do
+    before(:each) do
+      @default_admin_set = DefaultAdminSet.create! valid_attributes
```
bbpennel: Seems like `let(:default_admin_set) { DefaultAdminSet.create! valid_attributes }` should be in scope for the places where it is used below. Did you run into scoping issues?

spec/controllers/omniauth_controller_spec.rb
```diff
@@ -10,15 +10,19 @@
     end
 
     context 'not using database auth' do
-      it 'is successful' do
-        cached_database_auth = ENV['DATABASE_AUTH']
+      before do
+        @cached_database_auth = ENV['DATABASE_AUTH']
```
bbpennel: Would `let(:cached_database_auth) { ENV['DATABASE_AUTH'] }` work here? I think it would be in scope for the before and after.

spec/features/boxc_redirect_spec.rb
```diff
@@ -6,13 +6,19 @@
   cached_redirect_old_domain = ENV['REDIRECT_OLD_DOMAIN']
   cached_redirect_new_domain = ENV['REDIRECT_NEW_DOMAIN']
 
-  before(:each) do
+  before(:all) do
+    @tempfile = Tempfile.new('redirect_uuids.csv', 'spec/fixtures/')
```
bbpennel: can this be declared in a `let` statement or is the instance variable needed?

PR #473, opened 2019-11-18
Hyc 869 rake task tests

smithjp:
* https://jira.lib.unc.edu/browse/HYC-869


app/services/tasks/setup_tasks_service.rb
```diff
@@ -0,0 +1,45 @@
+module Tasks
```
bbpennel: Is this file only intended for use in running tests? If so, it would be helpful to indicate that in documentation, and maybe by putting this into a path that indicates this is a test helper. In longleaf I put test helper classes in `lib/longleaf/specs`, which I think I came up with after some googling. But I defer to where you think it best belongs

spec/services/tasks/proquest_ingest_service_spec.rb
```diff
@@ -0,0 +1,205 @@
...
+
+  describe '#migrate_proquest_packages' do
+    it 'ingests proquest records' do
+      allow(Date).to receive(:today).and_return(Date.parse('2019-09-12'))
+
+      expect{Tasks::ProquestIngestService.new(args).migrate_proquest_packages}.to change{ Dissertation.count }.by(7).and change{ DepositRecord.count }.by(7)
+
+      # check embargo information
+      dissertations = Dissertation.all
+      dissertation1 = dissertations[-7]
```
bbpennel: just curious, is there a reason to use negative indexes here instead of positive? It seems like the indexes of the variable names could line up with the indexes in the array

app/services/tasks/setup_tasks_service.rb
smithjp: No, some of these tasks are also used for setting up the VM.  I will add a note about the common usages of these tasks.

spec/services/tasks/proquest_ingest_service_spec.rb
smithjp: I am not currently clearing out other dissertation works from the test environment, so the ones ingested by the proquest script will be the last 7 and not necessarily the first 7.  I know there is at least one dissertation created in a feature test, and there might be another created elsewhere.  I can clear out all other dissertations and change this to have positive indices if that would be better.

bbpennel: Generally I'd prefer to avoid side effects between tests, since they can get pretty hard to manage. But would performing cleanup and repopulation significantly affect our already length test run time?

smithjp: It shouldn't slow things down too much, but I did realize that the proqest tests need to be refactored in general to test embargoes without creating works since they are causing travis-ci to time out.

spec/services/tasks/proquest_ingest_service_spec.rb
```diff
@@ -0,0 +1,355 @@
...
+                                                       deposit_package_type: 'http://proquest.com',
+                                                       deposit_package_subtype: nil,
+                                                       deposited_by: 'admin' })
+      expect(service.metadata_dir).to eq 'spec/fixtures/proquest'
+    end
+  end
+
+  describe '#migrate_proquest_packages' do
+    before do
+      Dissertation.delete_all
```
bbpennel: the fact that all it takes to wipe out all the dissertations is this slightly freaks me out, but we'll just have to be very careful with our scripts :)

PR #474, opened 2019-11-26
Allow less restrictive dates for general form

lfarrell:
* Allow less restrictive dates for general form
* Fix to have EDTF conversion work with multi-valued dates.

https://jira.lib.unc.edu/browse/HYC-861


lfarrell:
I'm slightly unclear as to what should be tested here, the multi_value_input input file? The added method is private. So, I'm not sure it can be tested. We could make it public. The edtf converter already has tests. Maybe the date issued edit field view file?

smithjp:
I was thinking of testing in the feature test, but it looks like the date format restrictions are not being applied in the tests, so I guess we can leave it untested until we think of a better way to test it.

PR #480, opened 2019-12-10
Selectively display "Remove" button for based_near field values

lfarrell:
https://jira.lib.unc.edu/browse/HYC-920


lfarrell:
Yeah, I think it's unrelated. I was seeing the same issue. It's doing it on master too.

lfarrell:
See https://github.com/UNC-Libraries/hy-c/pull/483.

PR #483, opened 2019-12-18
Selectively display "Remove" button for based_near field values

lfarrell:
https://jira.lib.unc.edu/browse/HYC-920
Supersedes https://github.com/UNC-Libraries/hy-c/pull/480, which had weird spacing on record view page that was unrelated to the pull request.

### DIFF
--- a/app/assets/javascripts/unc/unc_custom.js
+++ b/app/assets/javascripts/unc/unc_custom.js
@@ -102,12 +102,45 @@ $(function() {
         });
     }
 
+    // based_near is the only controlled field we use
+    // If we add more this function will need to be revised
+    function showRemoveOption() {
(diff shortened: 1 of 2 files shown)


smithjp:
If a person selects the wrong location and removes it, we need to have an empty box appear so that there is always at least 1 input box for location.  Right now I can select 1 location and then remove it without adding another location.

lfarrell:
Double checking, it looks like this pull request has the same behavior as master. On master, if you have one selected location and delete it you have to click the "Add another" link to have a new select box pop up. We can try and change it, but it seems more like a base Hyrax issue.

smithjp:
Ok, this seems like something we will want to fix in the future, then.  I have noticed some other problems with adding and removing input boxes in the forms too, so I will just write new tickets for those.

PR #489, opened 2020-01-15
Hyc 910 author order

smithjp:
* https://jira.lib.unc.edu/browse/HYC-910


app/renderers/hyrax/renderers/person_attribute_renderer.rb
```diff
@@ -1,36 +1,68 @@
...
-        if text =~ /ORCID:.*?http/
-          text_pieces = text.split(' ')
-          url = text_pieces[1].strip
-          text = "#{text_pieces[0]} <a href='#{url}' target='_blank'>#{url}</a>"
-        end
+      def render_dl_row
+        markup = ''
 
-        text
+        return markup if values.blank? && !options[:include_empty]
```
bbpennel: is `include_empty` every set to true?

app/renderers/hyrax/renderers/person_attribute_renderer.rb
```diff
@@ -1,36 +1,68 @@
...
+        markup << %(<dt>#{label}</dt>\n<dd><ul class='tabular'>)
+        attributes = microdata_object_attributes(field).merge(class: "attribute attribute-#{field}")
+        # conditional can be removed once all people objects have indexes
+        if values.first.match('index:')
+          Array(values.sort_by{|person| person.split('||').first.split(':').last}).each do |value|
```
bbpennel: It might make sense to implement a static `sort_by_index` method that you could use in the various places you're doing this. Its not totally intuitive at a glance what its doing with all the splitting, and is very formatting specific.

I'm actually wondering if it'd make sense to have a Person object class that could be initialized with strings in this format and provide accessor methods for the different parts of the name. Might also allow you to consolidate the pre-everything-indexed and the post-everything-indexed logic in one place. But maybe this is a step too far? Just seems like a pain doing splits all over the place.

spec/helpers/hyrax_helper_spec.rb
```diff
@@ -61,3 +61,5 @@
...
...ment of Biology", "orcid"=>"http://orcid.org/creator", "other_affiliation"=>"UNC"}, "1"=>{"name"=>"", "index"=>2, "affiliation"=>"", "orcid"=>"", "other_affiliation"=>""}}, "project_directors_attributes"=>{"0"=>{"index"=>1, "name"=>"project director", "affiliation"=>"Department of Biology", "orcid"=>"project director orcid", "other_affiliation"=>"UNC"}, "1"=>{"index"=>2, "name"=>"", "affiliation"=>"", "orcid"=>"", "other_affiliation"=>""}}, "researchers_attributes"=>{"0"=>{"index"=>1, "name"=>"researcher", "affiliation"=>"Department of Biology", "orcid"=>"researcher orcid", "other_affiliation"=>"UNC"}, "1"=>{"index"=>2, "name"=>"", "affiliation"=>"", "orcid"=>"", "other_affiliation"=>""}}, "reviewers_attributes"=>{"0"=>{"index"=>1, "name"=>"reviewer", "affiliation"=>"Department of Biology", "orcid"=>"reviewer orcid", "other_affiliation"=>"UNC"}, "1"=>{"index"=>2, "name"=>"", "affiliation"=>"", "orcid"=>"", "other_affiliation"=>""}}, "translators_attributes"=>{"0"=>{"index"=>1, "name"=>"translator", "affiliation"=>"Department of Biology", "orcid"=>"translator orcid", "other_affiliation"=>"UNC"}, "1"=>{"index"=>2, "name"=>"", "affiliation"=>"", "orcid"=>"", "other_affiliation"=>""}}}
```
bbpennel: Is this doing anything?

bbpennel:
Some other questions/observations:
* First advisor does not require affiliation, all the ones after require it
* Should the index field be visible? Its the only way to change order right now, so might be necessary
* The index of the Index field is always displayed as "Index (Creator #1)", "Index (Advisor #1)", etc.
* Sort of too bad that the author entries aren't in order in the edit view, but not essential

app/renderers/hyrax/renderers/person_attribute_renderer.rb
smithjp: I don't think we have it set to true anywhere, but this code came from the hyrax gem, so I wanted to leave it as close to the original as possible.

spec/helpers/hyrax_helper_spec.rb
smithjp: No, and I'm not really sure what it was from...  I'll remove it.

smithjp:
When the index field was discussed, we decided to leave it as an admin-only field for now, but we can always change it to be visible to everyone later.

PR #492, opened 2020-02-19
Update department service to find active term for id if it exists

smithjp:
* https://jira.lib.unc.edu/browse/HYC-956
* need to find active term for affiliations from foxml, which may be out of date

### DIFF
--- a/app/services/departments_service.rb
+++ b/app/services/departments_service.rb
@@ -19,17 +19,22 @@ module DepartmentsService
 
   def self.label(id)
     begin
+      term = authority.find(id).fetch('term')
       active = authority.find(id).fetch('active')
-      label = authority.find(id).fetch('term')
-      new_label = ''
-      while !active && new_label != label
-        if !new_label.blank?
-          label = new_label
+      if !active
+        next_term_exists = true
+        while next_term_exists
+          # check for next id
+          next_term = authority.all.select{ |item| item['id'] == term }
+          # return inactive label if no match
+          if next_term.blank?
+            next_term_exists = false
+          else
+            term = next_term.first['label']
+          end
         end
-        active = authority.find(label).fetch('active')
-        new_label = authority.find(label).fetch('term')
       end
-      label
+      term
     rescue
       Rails.logger.warn "DepartmentsService: cannot find '#{id}'"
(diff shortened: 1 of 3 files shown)


app/services/departments_service.rb
```diff
@@ -19,7 +19,22 @@ def self.identifier(term)
...
   def self.label(id)
     begin
-      authority.find(id).fetch('term')
+      term = authority.find(id).fetch('term')
+      active = authority.find(id).fetch('active')
+      if !active
+        next_term_exists = true
+        while next_term_exists
+          # check for next id
+          next_term = authority.all.select{ |item| item['id'] == term }
```
bbpennel: I think I understand that this is doing a walk through the vocab until it hits an active entry where the id matches the term of the previous inactive entry. But I'm not sure it works for hierarchical departmental names? For example:
```
  - id: BRIC
    term: School of Medicine; Biomedical Research Imaging Center
    active: false
  - id: Biomedical Research Imaging Center
    term: School of Medicine; Biomedical Research Imaging Center
    active: true
```
The term for BRIC is "School of Medicine; Biomedical Research Imaging Center" but the id for the active entry is just "Biomedical Research Imaging Center". Unless there is some splitting going on in the background, it seems like you would want to either check for a match on the last segment of the term of the inactive dept, or find other entries that match on the whole term?

Also not sure why `authority.all.select` when doing a search by id equality, doesn't `authority.find(id)` do that?

app/services/departments_service.rb
```diff
@@ -19,7 +19,22 @@ def self.identifier(term)
 
   def self.label(id)
     begin
-      authority.find(id).fetch('term')
+      term = authority.find(id).fetch('term')
+      active = authority.find(id).fetch('active')
```
bbpennel: Minor, but it'd probably be better to only do the find once, storing the result to a variable.

smithjp:
I just realized that this is not necessary.  I thought there was a problem with the service, but some terms which were recently inactivated were just not updated correctly in the yaml file.  The inactive terms are supposed to be updated to the new full affiliation term.  I'll close this pr and open a new one with the correct updates.

PR #497, opened 2020-03-10
1science script updates

smithjp:
* https://jira.lib.unc.edu/browse/HYC-963


app/services/tasks/onescience_ingest_service.rb
```diff
@@ -276,23 +280,148 @@ def parse_onescience_metadata(onescience_data)
...
+            end
+
+            # get affiliation info
+            organization = author_group.xpath('affiliation/organization').map(&:text).first
+            affiliation_id = author_group.xpath('affiliation/afid').text
+            department_id = author_group.xpath('affiliation/dptid').text
+
+            # add affiliation info to each person in group
+            if !organization.blank? && (!affiliation_id.blank? || !department_id.blank?)
+              author_group.xpath('.//author[not(@*)]').each do |author|
```
bbpennel: Do we not want to capture any authors with attributes? I don't know what the attributes are, just asking. Might be worth documenting somewhere.

app/services/tasks/onescience_ingest_service.rb
```diff
@@ -276,23 +280,148 @@ def parse_onescience_metadata(onescience_data)
...
+            author_id = author.xpath('auid').text
+            affiliations = record_affiliation_hash[author_id]
+
+            # split unc from external affiliations
+            unc_organizations = []
+            other_organizations = []
+            orcid = nil
+            affiliations.each do |affiliation|
+              orcid = affiliation['orcid']
+              if affiliation['afid'].match('60025111') ||
```
bbpennel: A short comment stating what all these IDs are would be helpful, or doing the matching in a well named helper method

app/services/tasks/onescience_ingest_service.rb
```diff
@@ -276,23 +280,148 @@ def parse_onescience_metadata(onescience_data)
...
+                  affiliation['afid'].match('60020469') ||
+                  affiliation['afid'].match('60072681') ||
+                  affiliation['afid'].match('113885172')
+                mapped_affiliation = mapped_affiliations.find{|mapped_dept| (mapped_dept['affiliation_id'] == affiliation['afid'] && mapped_dept['department_id'] == affiliation['dptid'])}
+                if mapped_affiliation
```
bbpennel: Could you add a comment here? I'm not really understanding why the organization goes in as a UNC organization because of the affiliation/dept_id not matching the vocab. Does this mean its a root affiliation, something like the whole university?

app/services/tasks/onescience_ingest_service.rb
```diff
@@ -276,23 +280,148 @@ def parse_onescience_metadata(onescience_data)
...
+                  unc_organizations << mapped_affiliation['mapped_affiliation']
+                else
+                  unc_organizations << affiliation['organization']
+                end
+              else
+                if affiliation['organization'].match('UNC')
```
bbpennel: Is this for tracking down affiliations that should possibly be in the affiliation mapping? would it make sense to add some context to the `puts` statement?

app/services/tasks/onescience_ingest_service.rb
```diff
@@ -276,23 +280,148 @@ def parse_onescience_metadata(onescience_data)
...
+                  puts author_id, affiliations
+                end
+                other_organizations << affiliation['organization']
+              end
+            end
+
+            if unc_organizations.count > 1
```
bbpennel: looks like the two cases here are identifical except for the second field being true or false. For the sake of drying things up, maybe only have one append block and use a variable for the second column.

spec/fixtures/onescience/onescience_config.yml
```diff
@@ -18,3 +18,6 @@ deposit_method: rake task
 deposit_type: a type
 deposit_subtype: a subtype
 deposit_record_id_log: spec/fixtures/onescience/1science_deposit_record_id.log
+scopus_xml_file: scopus-1-science-abstract.xml
```
bbpennel: wow, 358888 line fixture file? That probably doubles the size of the hy-c repo :) Is this an entire dump? Just wondering if it'd make sense to use a reduced set for testing. Might run faster too.

app/services/tasks/onescience_ingest_service.rb
```diff
@@ -276,23 +280,148 @@ def parse_onescience_metadata(onescience_data)
...
-        affiliations = affiliation_data['affiliation_author'+index.to_s]
-        people[index-1] = { 'name' => name,
-                            'orcid' => affiliation_data['ORCID_author'+index.to_s],
-                            'affiliation' => (affiliations.split('||') if !affiliations.blank?),
-                            'index' => index}
+      if !@scopus_hash[doi].blank?
+        people = @scopus_hash[doi]
+      else
+        affiliation_data = @affiliation_mapping.find{ |e| e['onescience_id'] == onescience_id }
+        (1..32).each do |index|
```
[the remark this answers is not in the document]
smithjp: That is the max number of author columns in the 1science spreadsheet.  I'll add a comment in the code.

app/services/tasks/onescience_ingest_service.rb
smithjp: good point.  they should all have affiliation data, but if not, we still want to check for orcids.

app/services/tasks/onescience_ingest_service.rb
```diff
@@ -328,28 +329,29 @@ def scopus_data
...
-
             # get affiliation info
-            organization = author_group.xpath('affiliation/organization').map(&:text).first
+            organizations = author_group.xpath('affiliation/organization').map(&:text)
             affiliation_id = author_group.xpath('affiliation/afid').text
             department_id = author_group.xpath('affiliation/dptid').text
 
             # add affiliation info to each person in group
-            if !organization.blank? && (!affiliation_id.blank? || !department_id.blank?)
               author_group.xpath('.//author[not(@*)]').each do |author|
```
bbpennel: This stuff isn't in an `if` anymore, so need to un-indent the block.

app/services/tasks/onescience_ingest_service.rb
```diff
@@ -370,40 +372,46 @@ def scopus_data
...
-                  affiliation['afid'].match('113885172')
+              # find all unc affiliations by scopus afid
+              # scopus afid 60025111 = The University of North Carolina at Chapel Hill
+              # scopus afid 60020469 = UNC School of Medicine
+              # scopus afid 60072681 = UNC Project-Malawi
+              # scopus afid 113885172 = UNC Project-China
+              # scopus afid 60013450 = UNC school of dentistry
+              # scopus afid 60005053 = Carolina Population Center
+              # scopus afid 60122501 = UNC business school
+              if (%w'60025111 60020469 60072681 113885172 60013450 60005053 60122501').include?(affiliation['afid'])
```
bbpennel: Might be good to make this array a constant for the class rather than defined inline here.

smithjp:
I am finally done making changes, and have tested the script on the test server

PR #501, opened 2020-05-18
Hyc 993 ingest oa books

smithjp:
* https://jira.lib.unc.edu/browse/HYC-993


app/services/tasks/csv_ingest_service.rb
```diff
@@ -0,0 +1,160 @@
...
+            attached_file_count = 0
+
+            files.each_with_index do |filename,file_index|
+              puts "[#{Time.now}] #{row['source_identifier']} attaching file #{file_index+1} of #{file_count}"
+
+              file_visibility = @config['file_visibility']
+
+              # parse filename
+              if filename.include? 'EPUB'
+                filename_with_extension = "#{filename}.epub"
```
bbpennel, line 91: just to be sure i understand, if the csv has a `file` field containing `9781469656489_EPUB`, the actual provided file is `9781469656489_EPUB.epub`?

app/services/tasks/csv_ingest_service.rb
```diff
@@ -0,0 +1,160 @@
...
+      end
+
+      puts "[#{Time.now}] Completed ingest of #{@config['batch_name']} in #{@config['metadata_file']}"
+    end
+
+
+    private
+
+      def create_deposit_record
+        if File.exist?(@config['deposit_record_id_log']) && !(File.open(@config['deposit_record_id_log']) {|f| f.readline}).blank?
```
bbpennel, line 137: the second part here is just checking if the log is empty, right? Could you use `File.zero?`, or is the concern that it might might contain whitespace? Just seems rather verbose at the moment, and the file gets read twice

app/services/tasks/csv_ingest_service.rb
smithjp: yes.  that is what the file names look like.

app/services/tasks/csv_ingest_service.rb
smithjp: i think i should be able to use File.zero. i'll try switching it.

PR #510, opened 2020-06-29
Add thumbnail id to oai feed

smithjp:


smithjp:
waiting on approval from RCMT

PR #517, opened 2020-06-30
bulkrax statuses

smithjp:
* https://jira.lib.unc.edu/browse/HYC-1018


app/jobs/bulkrax/import_work_job.rb
```diff
@@ -0,0 +1,31 @@
+# [hyc-override] use entry error information to determine if job has succeeded instead of build result
```
bbpennel, line 1: Is this something we could share back to the main project?

smithjp: I think this would be a good thing to at least ask about.  I'll add it to my list and start opening issues/prs next week.

PR #537, opened 2020-08-14
Remove index prefix from names in oai feed

smithjp:
* https://jira.lib.unc.edu/browse/HYC-1047

### DIFF
--- a/app/models/concerns/blacklight/document/dublin_core.rb
+++ b/app/models/concerns/blacklight/document/dublin_core.rb
@@ -38,7 +38,7 @@ module Blacklight::Document::DublinCore
         array_of_values = []
         if Array.wrap(values).first.match('index:')
           array_of_values = sort_people_by_index(values)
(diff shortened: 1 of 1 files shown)


app/models/concerns/blacklight/document/dublin_core.rb
```diff
@@ -38,7 +38,7 @@ def export_as_oai_dc_xml
         array_of_values = []
         if Array.wrap(values).first.match('index:')
           array_of_values = sort_people_by_index(values)
-          array_of_values.map!{|value| value.gsub(/\Aindex:\d+\|\|/, '')}
+          array_of_values.map!{|value| value.gsub(/\Aindex:\d+\|\|/, '').gsub(/\Aindex:\|\|/, '')}
```
lfarrell, line 41: Do you think this would work `value.gsub(/\Aindex:(\d+)?\|\|/, '')`, making the digits optional instead having to have a double gsub?

smithjp: you're right, it would be better to consolidate.  thanks!

PR #567, opened 2020-10-12
Add methods to base actor to apply work-specific workflow permissions

smithjp:
* https://jira.lib.unc.edu/browse/HYC-992

### DIFF
--- a/app/actors/hyrax/actors/base_actor.rb
+++ b/app/actors/hyrax/actors/base_actor.rb
@@ -140,7 +140,7 @@ module Hyrax
       # [hyc-override] added this method to allow work-specific permissions to work
       def apply_work_specific_permissions(env)
         permissions_attributes = env.attributes['permissions_attributes']
-        return if permissions_attributes.blank?
+        return true if permissions_attributes.blank?
         workflow = Sipity::Workflow.where(permission_template_id: env.curation_concern.admin_set.permission_template.id,
                                           active: true).first
         entity = Sipity::Entity.where(proxy_for_global_id: env.curation_concern.to_global_id.to_s, workflow_id: workflow.id).first_or_create!
--- a/app/actors/hyrax/actors/base_actor.rb
+++ b/app/actors/hyrax/actors/base_actor.rb
@@ -20,9 +20,7 @@ module Hyrax
       def create(env)
         apply_creation_data_to_curation_concern(env)
         apply_save_data_to_curation_concern(env)
-        save(env)
-        apply_work_specific_permissions(env)
-        next_actor.create(env) && run_callbacks(:after_create_concern, env)
+        save(env) && next_actor.create(env) && run_callbacks(:after_create_concern, env) && apply_work_specific_permissions(env)
       end
 
       # @param [Hyrax::Actors::Environment] env
@@ -91,6 +89,8 @@ module Hyrax
       # TODO this method could move to the work form.
       def clean_attributes(attributes)
         attributes[:license] = Array(attributes[:license]) if attributes.key? :license
+        # [hyc-override] Overriding actor to cast rights statements as single valued
+        # removed rights_statement-specific line of code so that it could be cast in `remove_blank_attributes!`
         # [hyc-override] remove index field if for some reason is added to permissions_attributes hashes
         if !attributes['permissions_attributes'].blank?
           permission_attrs = {}
@@ -103,7 +103,6 @@ module Hyrax
           end
           attributes['permissions_attributes'] = permission_attrs
         end
-        # [hyc-override] Overriding actor to cast rights statements as single valued
         remove_blank_attributes!(attributes)
       end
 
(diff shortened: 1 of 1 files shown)


app/actors/hyrax/actors/base_actor.rb
```diff
@@ -87,6 +91,18 @@ def apply_save_data_to_curation_concern(env)
       # TODO this method could move to the work form.
       def clean_attributes(attributes)
         attributes[:license] = Array(attributes[:license]) if attributes.key? :license
+        # [hyc-override] remove index field if for some reason is added to permissions_attributes hashes
+        if !attributes['permissions_attributes'].blank?
+          permission_attrs = {}
+          attributes['permissions_attributes'].each do |k,v|
+            if !v['index'].blank?
```
bbpennel, line 98: is the presence of the `index` field new in this PR or has it been there a while, you're just taking this opportunity to clean it up since you're already doing stuff with permissions? Which is fine, I'm just trying to understand if its possible to prevent it from appearing in the first place, or if its better to clean it up afterwards.

app/actors/hyrax/actors/base_actor.rb
```diff
@@ -19,7 +20,9 @@ class BaseActor < AbstractActor
       def create(env)
         apply_creation_data_to_curation_concern(env)
         apply_save_data_to_curation_concern(env)
-        save(env) && next_actor.create(env) && run_callbacks(:after_create_concern, env)
+        save(env)
+        apply_work_specific_permissions(env)
+        next_actor.create(env) && run_callbacks(:after_create_concern, env)
```
bbpennel, line 25: Will it cause any issues that the `save` method is no longer a part of the return value for this method?

app/actors/hyrax/actors/base_actor.rb
```diff
@@ -121,6 +137,44 @@ def log_deleted_people_objects(attributes, work_id)
...
+
+      # added this method to allow work-specific permissions to work
+      def apply_work_specific_permissions(env)
+        permissions_attributes = env.attributes['permissions_attributes']
+        return if permissions_attributes.blank?
+        workflow = Sipity::Workflow.where(permission_template_id: env.curation_concern.admin_set.permission_template.id,
+                                          active: true).first
+        entity = Sipity::Entity.where(proxy_for_global_id: env.curation_concern.to_global_id.to_s).first
+        permissions_attributes.each do |k,permission|
+          if !permission['id'].blank?
```
bbpennel, line 149: could you add a comment describing why the presence of an id for a permission means that it doesn't need to add in the work-specifc workflow details?

app/actors/hyrax/actors/base_actor.rb
```diff
@@ -121,6 +137,44 @@ def log_deleted_people_objects(attributes, work_id)
           end
         end
       end
+
+      # added this method to allow work-specific permissions to work
```
bbpennel, line 141: maybe add `[hyc-override]` to this comment?

app/actors/hyrax/actors/base_actor.rb
```diff
@@ -121,6 +137,44 @@ def log_deleted_people_objects(attributes, work_id)
...
+            roles = 'approving'
+          else
+            roles = 'viewing'
+          end
+
+          create_workflow_permissions(entity, agents, roles, workflow)
+        end
+      end
+
+      # added this method to allow work-specific permissions to work
```
bbpennel, line 171: maybe add `[hyc-override]` to this comment?

app/actors/hyrax/actors/base_actor.rb
```diff
@@ -87,6 +91,18 @@ def apply_save_data_to_curation_concern(env)
...
+          attributes['permissions_attributes'].each do |k,v|
+            if !v['index'].blank?
+              permission_attrs[k] = v.except('index')
+            else
+              permission_attrs[k] = v
+            end
+          end
+          attributes['permissions_attributes'] = permission_attrs
+        end
         # [hyc-override] Overriding actor to cast rights statements as single valued
```
bbpennel, line 106: i know its not part of this PR, but should this comment be above line 93?

bbpennel (changes requested):

app/actors/hyrax/actors/base_actor.rb
smithjp: This is something I found while working on this ticket, and I had to fix it in order to get things working.  I'm not sure where the index param is coming from.

app/actors/hyrax/actors/base_actor.rb
smithjp: yes, i wasn't sure what that was.  it appears to be a comment that was supposed to indicate that a line of code was removed.  i will add a comment.

bbpennel: ah, so its a removed line? I wasn't sure if it was referring to `attributes[:license] = Array(attributes[:license]) if attributes.key? :license`

PR #568, opened 2020-10-12
Use number of records as row limit in solr queries

smithjp:
* https://jira.lib.unc.edu/browse/HYC-1060
* also will open PR in gem

### DIFF
--- a/app/parsers/bulkrax/csv_parser.rb
+++ b/app/parsers/bulkrax/csv_parser.rb
@@ -157,7 +157,8 @@ module Bulkrax
     end
 
     def create_from_collection
-      work_ids = ActiveFedora::SolrService.query("member_of_collection_ids_ssim:#{importerexporter.export_source}").map(&:id)
+      collection_works_count = ActiveFedora::SolrService.get("member_of_collection_ids_ssim:#{importerexporter.export_source}", rows: 0)['response']['numFound']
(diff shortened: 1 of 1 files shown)


app/parsers/bulkrax/csv_parser.rb
```diff
@@ -157,7 +157,8 @@ def create_from_importer
     end
 
     def create_from_collection
-      work_ids = ActiveFedora::SolrService.query("member_of_collection_ids_ssim:#{importerexporter.export_source}").map(&:id)
+      collection_works_count = ActiveFedora::SolrService.get("member_of_collection_ids_ssim:#{importerexporter.export_source}", rows: 0)['response']['numFound']
```
bbpennel, line 160: if the objective is to always return all rows in the next query, it might make sense to save a round trip to solr and simply set `rows` in the next query to a super high number

app/parsers/bulkrax/csv_parser.rb
```diff
@@ -166,7 +167,8 @@ def create_from_collection
     end
 
     def create_from_worktype
-      work_ids = ActiveFedora::SolrService.query("has_model_ssim:#{importerexporter.export_source}").map(&:id)
+      work_type_count = ActiveFedora::SolrService.get("has_model_ssim:#{importerexporter.export_source}", rows: 0)['response']['numFound']
```
bbpennel, line 170: same question/suggestion as the other method

bbpennel (approved):
Looks good, has it been deployed to a server for rebekah or deseree to test yet?

PR #569, opened 2020-10-13
bulkrax parsing

smithjp:
* https://jira.lib.unc.edu/browse/HYC-1066


bbpennel (approved):
I'm sort of wondering about the safety of splitting on anything that might show up in normal user input (`;` seems likely), but if rebekah is okay with these delimiters than this all looks good to me

PR #580, opened 2020-11-06
Add a creation date for 1science works during ingestion

lfarrell:
https://jira.lib.unc.edu/browse/HYC-1082

### DIFF
--- a/app/services/tasks/onescience_ingest_service.rb
+++ b/app/services/tasks/onescience_ingest_service.rb
@@ -154,7 +154,6 @@ module Tasks
 
               # create and save file
               file_attributes = { title: [filename],
-                                  date_created: Time.new.strftime("%Y-%m-%dT%H:%M:%S+00:00"),
                                   related_url: [source_url] }
               file_set = FileSet.create(file_attributes)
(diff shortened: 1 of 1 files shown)


app/services/tasks/onescience_ingest_service.rb
```diff
@@ -154,7 +154,7 @@ def ingest
 
               # create and save file
               file_attributes = { title: [filename],
-                                  date_created: work_attributes['date_issued'],
+                                  date_created: Time.new.strftime("%Y-%m-%dT%H:%M:%S+00:00"),
```
bbpennel, line 157: Based off of this request from Anna:
> It looks like both Date Created and Date Uploaded/Submitted were added to the metadata, is it possible to only have Date Uploaded/Submitted added?

Would it work if we didn't set `date_created` at all, and instead set `date_uploaded`? It looks like that field is displayed if `date_created` isn't set in the view that I believe is generating the table of filesets:
https://github.com/UNC-Libraries/hy-c/blob/main/app/views/hyrax/base/_member.html.erb#L7

lfarrell:
This PR was rolled into https://github.com/UNC-Libraries/hy-c/pull/651, which has been merged

PR #588, opened 2020-11-17
Set a max height for thumbnails

lfarrell:
Related to https://jira.lib.unc.edu/browse/HYC-1075


bbpennel:
Going to hold off on these changes til we spend more time figuring out cause of the weird thumbnails

PR #607, opened 2021-02-18
Gh actions

lfarrell:
https://jira.lib.unc.edu/browse/HYC-865


.github/workflows/build.yml
```diff
@@ -0,0 +1,115 @@
...
+    - name: Update rubygems
+      run: |
+        gem update --system
+        gem install bundler:2.1.4
+
+    - name: Set up JDK 1.8
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
+    - name: Cache Maven packages
```
bbpennel, line 76: Is this needed? I don't think anything is being build with maven here.

bbpennel (changes requested):

.github/workflows/build.yml
lfarrell: Yeah, probably not.

bbpennel (approved):

PR #625, opened 2021-04-29
Only convert input field dates to edtf

lfarrell:
https://jira.lib.unc.edu/browse/HYC-1122
https://jira.lib.unc.edu/browse/HYC-1141


bbpennel (approved):
This look good. We'll need to do some QA of the forms to make sure there aren't any unexpected side effects

PR #634, opened 2021-05-05
Check users exist before assigning email recipients

lfarrell:
https://jira.lib.unc.edu/browse/HYC-1131

### DIFF
--- a/app/jobs/assign_permissions_to_work_job.rb
+++ b/app/jobs/assign_permissions_to_work_job.rb
@@ -11,10 +11,10 @@ class AssignPermissionsToWorkJob < Hyrax::ApplicationJob
     entity = Sipity::Entity.where(proxy_for_global_id: work.to_global_id.to_s).first
     recipients = Hash.new
     selected_role = Role.where(name: group_name).first
-    if selected_role.respond_to?(:users)
+    if !selected_role.nil?
       recipients[:to] = selected_role.users
     else
-      Rails.logger.warn "No users found for role: #{selected_role} on work: #{work_id} for group: #{group_name}"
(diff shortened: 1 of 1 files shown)


app/jobs/assign_permissions_to_work_job.rb
```diff
@@ -10,7 +10,13 @@ def perform(model, work_id, group_name, type, access)
     # Send notification to reviewer group
     entity = Sipity::Entity.where(proxy_for_global_id: work.to_global_id.to_s).first
     recipients = Hash.new
-    recipients[:to] = Role.where(name: group_name).first.users
+    selected_role = Role.where(name: group_name).first
+    if selected_role.respond_to?(:users)
```
bbpennel, line 14: This should work, although I sort of wonder if it would make more sense as a nil check? Is it likely that this query is going to give back and object that is not a role?

lfarrell: I'm inclined to leave it as it is since it was throwing a NoMethodError. Checking `Role.where(name: group_name).first.users.nil?` actually returns false if the method is undefined.

bbpennel: I think it is throwing "NoMethodError" because `Role.where(name: group_name).first` is returning `nil`, for instance:
```
irb(main):004:0> nil.users
(irb):4:in `<main>': undefined method `users' for nil:NilClass (NoMethodError)

irb(main):007:0> [].first.users
(irb):7:in `<main>': undefined method `users' for nil:NilClass (NoMethodError)
```
Although I'm not sure why calling `Role.where(name: group_name).first.users.nil?` would give false instead of throwing a NoMethodError.

PR #654, opened 2021-07-16
Add static method to virus checking

bbpennel:


spec/lib/hyc/virus_scanner_spec.rb
```diff
@@ -36,17 +31,12 @@
   context 'when a file is infected' do
     src_path = Pathname.new('spec/fixtures/files/virus.txt').realpath.to_s
 
-    if Dir.pwd.to_s.include? "runner"
-      let(:file) { Tempfile.new.path }
+    if ENV["CI"].nil?
```
bbpennel, line 34: I think this has to go in a `before` block, you may need to declare it as `@file`

PR #658, opened 2021-08-03
Catch blacklight sort and number formatting errors

lfarrell:
https://jira.lib.unc.edu/browse/HYC-1160
https://jira.lib.unc.edu/browse/HYC-1167

Depends on https://gitlab.lib.unc.edu/cdr/hyc-gems/-/merge_requests/64

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -40,9 +40,8 @@ class ApplicationController < ActionController::Base
     def render_rsolr_exceptions(exception)
       exception_text = exception.to_s
 
-      if exception_text.include?("java.lang.NumberFormatException")
-        render_500
-      elsif exception_text.include?("Can't determine a Sort Order")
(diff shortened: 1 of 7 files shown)


app/controllers/application_controller.rb
```diff
@@ -24,10 +24,31 @@ class ApplicationController < ActionController::Base
...
 
+    def render_500
+      render 'errors/internal_server_error', status: 500
+    end
+
+    # Error caught in catalogController
+    def render_rsolr_exceptions(exception)
+      exception_text = exception.to_s
+
+      if exception_text.include?("java.lang.NumberFormatException")
```
bbpennel, line 43: I think this one would be a 400 as well since it reflects bad user input as well

bbpennel (changes requested):
This looks like a good direction for handling these issues. Just one suggestion prior to testing related to the number format exception

bbpennel (approved):

bbpennel (approved):

PR #660, opened 2021-08-04
Skip applying work specific permissions for filesets

lfarrell:
File sets don't have admin sets. So updating them independently of their work should skip apply_work_specific_permissions update.
https://jira.lib.unc.edu/browse/HYC-1203

### DIFF
--- a/app/actors/hyrax/actors/base_actor.rb
+++ b/app/actors/hyrax/actors/base_actor.rb
@@ -141,30 +141,34 @@ module Hyrax
       def apply_work_specific_permissions(env)
         permissions_attributes = env.attributes['permissions_attributes']
         return true if permissions_attributes.blank?
-        workflow = Sipity::Workflow.where(permission_template_id: env.curation_concern.admin_set.permission_template.id,
-                                          active: true).first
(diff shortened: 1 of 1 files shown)


app/actors/hyrax/actors/base_actor.rb
```diff
@@ -141,30 +141,34 @@ def log_deleted_people_objects(attributes, work_id)
...
-          end
-          agents = [Sipity::Agent.where(proxy_for_id: agent_id, proxy_for_type: agent_type).first_or_create]
 
-          if permission['access'] == 'edit'
-            roles = 'approving'
-          else
-            roles = 'viewing'
-          end
+        # File sets don't have admin sets. So updating them independently of their work should skip this update.
+        if env.curation_concern.respond_to? :admin_set
```
bbpennel, line 146: Instead of doing this as an `if`, how about doing something like
```
return true unless env.curation_concern.respond_to? :admin_set
```
That won't require indenting everything, so will be a much smaller change

bbpennel (changes requested):
This looks like a good approach and sounds like it works, just one suggestion to reduce the size of the change

bbpennel (approved):

PR #665, opened 2021-08-17
Bulkrax and Rails updates

lfarrell:
https://jira.lib.unc.edu/browse/HYC-1209
Depends on https://gitlab.lib.unc.edu/cdr/hyc-gems/-/merge_requests/66


bbpennel (approved):
Looks good to me in so far as I can tell from reading, will mostly have to depend on testing to determine if things are in good shape. Seems like you did a good job of merging our overrides into updated files in the cases where I checked

PR #666, opened 2021-08-17
WIP: Rails 5.2 updates

lfarrell:
https://jira.lib.unc.edu/browse/HYC-1153

Depends on https://github.com/UNC-Libraries/hy-c/pull/665. Rails can't be updated until hyrax is updated.


lfarrell:
Combined with #665

PR #693, opened 2021-11-05
Don't hardcode file name and mime type if replacing a file

lfarrell:
- https://jira.lib.unc.edu/browse/HYC-1229
- https://jira.lib.unc.edu/browse/HYC-1268

### DIFF
--- a/app/models/concerns/bulkrax/file_factory.rb
+++ b/app/models/concerns/bulkrax/file_factory.rb
@@ -92,10 +92,10 @@ module Bulkrax
         fileset.files.first.create_version
         opts = {}
         opts[:path] = fileset.files.first.id.split('/', 2).last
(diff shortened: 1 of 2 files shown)


maxkadel (approved):

lfarrell:
I don't what this means :/ mwerp. Also, I'm not entirely sure what the set_removed_filesets method is supposed to accomplish; add a removed.png file and then create derivatives for it??? I'm inclined to disable the method.

bbpennel:
My recollection is that Rebekah wants the old versions of files to be marked as private. When testing this branch on the server, each new version of a file would result in a previous version of the file appearing to be retained (all the metadata stayed the same) but downloading it would give you a 0 byte file. So I think the behavior we want is the old version becoming private and NOT creating a new zeroed out version.

maxkadel (approved):

PR #701, opened 2021-11-09
Ensure tests don't leave artifacts in dev environment

maxkadel:


spec/features/boxc_redirect_spec.rb
```diff
@@ -2,23 +2,22 @@
...
       f.puts 'uuid,new_path'
     end
     ENV['REDIRECT_OLD_DOMAIN'] = ENV['HYRAX_HOST'].gsub('https://','')
     ENV['REDIRECT_NEW_DOMAIN'] = 'dcr-test.lib.unc.edu'
     stub_request(:any, 'https://dcr-test.lib.unc.edu/list/uuid:02fc897a-12b6-4b81-91e4-b5e29cb683a6').to_return(status: 200)
-  end
-
-  after(:all) do
+    # example.run is where the test actually runs
+    example.run
```
bbpennel, line 18: I haven't used around blocks before, could you explain why they are preferable in this case to using before and after blocks? Does the code after the example.run invocation still execute in the case of a test failure?

maxkadel: You can see more about them here - https://relishapp.com/rspec/rspec-core/v/3-10/docs/hooks/around-hooks 

The "after" parts still run even if a test fails. 

I'm not totally attached to using "around" rather than before and after hooks, but I wanted to make them a little more consistent, and be sure to set any environment variables back to what they were prior to the test. Part of why I opted for the around hook was so that we could scope the variables more conservatively.

bbpennel: Thanks, I did initially look at that link and was wondering what was considered the standard practice for this purpose because of this quote:
> Of course, you can do the same thing with before and after hooks; and it's often cleaner to do so.
so i wasn't sure which was considered
Making sure the environment variables get reset is important, and I'm fine with this usage if you feel like it is better in this case

bbpennel (approved):

PR #702, opened 2021-11-09
Revert dependabot shex update

maxkadel:
For currently undiagnosed reasons, the shex update brings the site down, despite the fact it's supposedly a patch update


maxkadel:
From apache logs 

```
[ E 2021-11-09 13:29:01.2692 7469/T1r age/Cor/App/Implementation.cpp:221 ]: Could not spawn process for application /htdocs/ir: The application encountered the following error: You have already activated logger 1.3.0, but your Gemfile requires logger 1.4.3. Since logger is a default gem, you can either remove your dependency on it or try updating to a newer version of bundler that supports logger as a default gem. (Gem::LoadError)
  Error ID: 766e37e6 
```

PR #707, opened 2021-11-18
Get derivatives test working locally

maxkadel:
- Still have not entirely tangled the temp directory situation, have separate ticket for that


maxkadel:
For now, we do not want to delete working files. May revisit this.

PR #710, opened 2021-11-22
Get test for CreateDerivativesJob passing on VM and in CI

maxkadel:
- The test currently assumes that the data and temp storage are in the same place. Add TODO to update the test to treat these directories separately
- In the context of the test, the virus checker method is not finding the file that's passed in, even though it is present. This does not seem to be an issue in production, and is not the focus of these tests, so mock out the virus check method

### DIFF
--- a/spec/fixtures/tmp/.keep
+++ b/spec/fixtures/tmp/.keep
@@ -1,0 +1,1 @@
+This should be the only file in this directory. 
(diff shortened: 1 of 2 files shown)


maxkadel:
PR opened against wrong branch

lfarrell (approved):

PR #712, opened 2021-11-23
OAI endpoint tests passing on VM

maxkadel:
- Ensure OAI request tests pass on the VM and in CI
- Will allow us to be more confident in test setup and setup tests faster
- Will eventually pull down on unintended test interactions as we move to running tests in different orders


spec/requests/oai_pmh_endpoint_spec.rb
```diff
@@ -7,21 +9,25 @@
...
   let(:document_config) { { limit: limit } }
   let(:oai_config) { { provider: provider_config, document: document_config } }
-  let(:timestamps) { Array.new(0) }
-
+  let(:timestamps) { solrRecords['response']['docs'].map { |doc| doc["timestamp"] }.sort! }
+  let(:solrRecords) do
+    ActiveFedora::SolrService.get('has_model_ssim:(Dissertation OR Article OR MastersPaper OR '+ 'HonorsThesis OR Journal OR DataSet OR Multimed OR ScholarlyWork ' + 'OR General OR Artwork) AND visibility_ssi:open', rows: 100)
+  end
+  before(:all) do
+    solr = Blacklight.default_index.connection
```
bbpennel, line 17: I assume with this methodology that if there are any changes to indexing we will need to represent them in the sample document? It may be valuable to document how you generated the file since we'll likely have to update it in the future.

Also, is there anything else in the index prior to this deletion aside from the sample documents? I half remember hyrax putting entries in for every fedora object, rather than just the works. Is there any chance that wiping the index will impact other tests, since I thought I remember you mentioning that there is the potential for modifications from one test affecting other tests?

spec/requests/oai_pmh_endpoint_spec.rb
```diff
@@ -57,6 +63,7 @@
       let(:document_config) { { limit: 25 } }
 
       scenario 'a resumption token is provided' do
+        # TODO: freeze time, so if the test is slow we won't get strange failures
```
bbpennel, line 66: While I love this comment and the special effects that go with it, seems like it'd make sense to go ahead and do that as part of this PR unless its impractical to do so?

spec/requests/oai_pmh_endpoint_spec.rb
maxkadel: Good call on documenting how I did it - it didn't make sense to me to totally script it, but I've put it in the readme, and if we decide we're doing it enough to need to script it we can do that. 

So, right now, we are running a rake task that runs all the tasks in the SetupTasksService, including importing test data, which creates fairly full-fledged objects in both Fedora and Solr. The problem is that the task is fairly slow, so for development you just run it once, then hope that your tests don't change any of the objects. 

I want to shift to a) running the tests in random order, so we raise any order dependencies before we hit them in production (there's a ticket in the backlog for this) and b) be much more explicit about what we can and cannot expect to exist in the context of any test, and do the minimal setup possible for that test. 

This takes us one step towards those things, and doesn't make the current test suite any more broken locally or in CI than it was before, so I think it's an improvement.

spec/requests/oai_pmh_endpoint_spec.rb
maxkadel: I put the timestamps in the solr documents, so we don't need this any more. Before, it was dependent on whenever the rake task had saved the work to solr, and that each of those saves happened on a different second, and if two were saved too close together some of the tests would fail. So, we're fixing that, and we don't need to freeze time! Although that can be a super handy tool if the test is explicitly about time passing. https://api.rubyonrails.org/v5.2.6/classes/ActiveSupport/Testing/TimeHelpers.html

README.md
```diff
@@ -18,5 +18,29 @@
 * Import scripts for new ProQuest deposits
 * Merges old and new analytics data from Google Analytics
 
+#### Testing
+* Creating Solr fixture objects (see `spec/support/oai_sample_solr_documents.rb` )
+  * Run the test_setup rake task in order to get all the fixtures in the development environment
+  ```
+  bundle exec rake test_setup RAILS_ENV=development
+  ```
+  * Search in your development environment front end to find the ID for the object you you want to update (if using the UNC-CH VM this will be at localhost:4040, if you are running it locally on the default rails server, it will be at localhost:3000)
```
bbpennel, line 27: Extra "you", should be: `ID for the object you want to update`

README.md
```diff
@@ -18,5 +18,29 @@
...
+    Example:
+    ```ruby
+    work = Article.find("5x21tf41f")
+    work.to_solr.deep_symbolize_keys!
+    ```
+    ```ruby
+    work = YourModel.find("YourIdAsAString")
+    work.to_solr.deep_symbolize_keys!
+    ```
+    * Copy the output of this to the sample_solr_documents file. You may need to add a `:timestamp` in order for the `spec/requests/oai_pmh_endpoint_spec.rb` test to continue to pass.
```
bbpennel, line 43: Please state under what circumstance you would need to add a `:timestamp`

bbpennel (changes requested):
Two minor suggestions

bbpennel (approved):

PR #715, opened 2021-11-24
Idempotent tests

maxkadel:
Overall goal is to be able to run tests on VM and have them consistently pass and not change development environment or be affected by changes to development environment.

- For some tests, clean out Fedora, Solr, and database prior to test, to ensure we can create a new unique object when we expect to.
- Create object in test, rather that prior to all tests. 
- Ensure consistent environment for tests
- Delete temp files after test completes
- Mark test that is flaky on VM (both on main and this branch) as pending
  - Have created ticket to deprecate or refactor this code - https://jira.lib.unc.edu/browse/HYC-1273


maxkadel:
Replaced with https://github.com/UNC-Libraries/hy-c/pull/717 to clean up history

PR #718, opened 2021-11-29
Add linter, minimal changes to run in CI

maxkadel:
- Add Bixby gem (Samvera code linter based on Rubocop)
- Fix only problems leading to Rubocop errors, adding all other to rubocop_todo

Connected to https://jira.lib.unc.edu/browse/HYC-1256 
See also https://gitlab.lib.unc.edu/cdr/hyc-gems/-/merge_requests/77


app/services/tasks/proquest_ingest_service.rb
```diff
@@ -55,7 +55,7 @@ def migrate_proquest_packages
         end
         pdf_file = Dir.glob("#{unzipped_package_dir}/*.pdf")
         if pdf_file.count == 1
-          pdf_file = pdf_file.first.to_s
+          pdf_file.first.to_s
```
bbpennel, line 58: I may just not be familiar enough with ruby, but it seems like this line is no longer having any effect? The `if` statement is not part of the assignment on line 56,so it seems like pdf_file is no longer getting converted from an array to a single entry.

app/models/file_download_stat.rb
```diff
@@ -1,3 +1,5 @@
+# # TODO: See if this file is actually being used / picked up. Based on its location, and the fact that all tests pass
```
bbpennel, line 1: Looking at the history of this file, it seems like it must have been significant at some point for getting download stats working. It might just not be getting tested?

app/models/file_download_stat.rb
```diff
@@ -33,14 +35,13 @@ def ga_statistics(start_date, file)
...
-      filter_id = file.id
+      redirect_path = redirect_lookup('new_path', file.id)
 
-      # check if file was migrated
-      redirect_path = redirect_lookup('new_path', filter_id)
-
-      if redirect_path
-        filter_id = "#{filter_id}|#{redirect_path['uuid']}"
-      end
+      _filter_id = if redirect_path
```
bbpennel, line 40: While this syntax looks unexpected to me as a primarily java developer, I don't have any problem with it. I was wondering, is this something rubocop is enforcing? Also, is the indent of the subsequent lines supposed to be a standard indent size or should the indent of the `else`/`end` line up with the `if`? They do for the first one of these blocks at line 23, so i wasn't sure.

maxkadel: The underscore before the variable name indicates that it isn't being used (and Rubocop does enforce this). Because the file doesn't seem to be tested, I didn't want to change it too much, so I left the conditional in there and just made it clear it's not currently being used.

Fixing the indent.

bbpennel: Ah, I didn't realize that starting a variable name with an underscore had that extra meaning, good to know.

PR #722, opened 2021-12-03
Add version footer

maxkadel:
Connected to https://jira.lib.unc.edu/browse/HYC-1264 

- Styling to maintain look and feel
- Set constants in an initializer (the pattern that DCE has used - makes it harder to test production-only code, though)

CDR footer on production in v1.0.0:
image

Deployed to cdr-test  - mobile
image

Deployed to cdr-test - desktop
image

### DIFF
--- a/app/assets/stylesheets/unc_custom.css.scss
+++ b/app/assets/stylesheets/unc_custom.css.scss
@@ -347,6 +347,10 @@ div.person {
       }
     }
   }
+
+  #unc-version-footer {
+    margin-top: 15px;
+  }
 }
 
 div.unc-modal {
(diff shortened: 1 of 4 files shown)


bbpennel:
I know its from the DCE exampe, but I'm not sure what "not in deployed environment" means, is that to say its on in a production env? Is that needed because there isn't a timestamp available, or just to make it clear what environment you're in? 

I also sort of wonder if 2 lines in the footer is a lot of vertical space to dedicate to info about the software version, but I'd leave that up to rebekah to decide on. I also haven't looked at it on a mobile device

maxkadel:
"Not in deployed environment" means rails is running in test or development mode, and not deployed to a server. 

Just checked it on an actual mobile phone (used the dev tools before), and the "stick to bottom" is making the the footer worse on mobile. Do we know why we're not turning it into a "hamburger" menu? Is that some kind of requirement?

maxkadel (commented):
I can't submit a review, since I opened this PR, but with Dean's changes this is looking good.

PR #725, opened 2021-12-07
Send referrer to Google analytics

maxkadel:
- Test what we are sending to GA
- Simplify conditional
- Use event builder, since it is easier to inspect in debugger

With log level :debug on cdr-qa: 
```
2021-12-08 12:44:12 -0500 - DEBUG: [40c764d0-8230-4d56-943b-50ea63df8eb2] DownloadAnalyticsBehavior request.referrer: https://cdr-qa.lib.unc.edu/concern/multimeds/qf85nb60x?locale=en
```

Connected to https://jira.lib.unc.edu/browse/HYC-1287


app/controllers/concerns/hyc/download_analytics_behavior.rb
```diff
@@ -6,23 +6,29 @@ module DownloadAnalyticsBehavior
...
-                          user_agent: request.headers['User-Agent'],
-                          user_ip: request.remote_ip)
-            tracker.track
-          end
+        if Hyrax.config.google_analytics_id.present? && !request.url.match('thumbnail')
+          # Staccato works with Google Analytics v1 api:
+          # https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters
+          # Staccato on Github: https://github.com/tpitale/staccato
+          request_referrer = request.referrer
+          medium = request_referrer.present? ? 'referral' : 'direct'
```
bbpennel, line 14: Under what circumstances were you finding referrer not to be set, or to be set? Is it mostly when using a direct link and/or client like curl? Just trying to get a sense of the value of capturing the distinction

bbpennel (approved):

app/controllers/concerns/hyc/download_analytics_behavior.rb
maxkadel: Yes - right now they're all being set to direct, and it looks like, at least with the real time API, it will only show the referrer if the medium is set to "referral", but I didn't want to just always automatically set it to referral in case that messed up the downstream data too much.

PR #728, opened 2021-12-10
Sage Ingest rake task - unpack files and log to start with

maxkadel:
Connected to https://jira.lib.unc.edu/browse/HYC-1281

Start rake task to ingest packages provided by Sage

Thus far, this rake task should:
* Take a path to a config file 
* Log progress and errors to the rails log 
* Unpack the zip file 
* Be aware of where the unzipped xml and pdf files are 
* Write to the progress log for the ingest when the ingest is "done" (the concept of "done" will evolve as the ingest evolves)


app/services/tasks/sage_ingest_service.rb
```diff
@@ -0,0 +1,83 @@
...
+      @unzip_dir = config['unzip_dir']
+    end
+
+    def process_packages
+      sage_package_paths = Dir.glob("#{@package_dir}/*.zip").sort
+      count = sage_package_paths.count
+      Rails.logger.tagged('Sage ingest') { Rails.logger.info("Beginning ingest of #{count} Sage packages") }
+      sage_package_paths.each.with_index(1) do |package_path, index|
+        Rails.logger.tagged('Sage ingest') { Rails.logger.info("Begin processing #{package_path} (#{index} of #{count})") }
+        orig_file_name = package_path.split('.zip')[0].split('/')[-1]
```
bbpennel, line 18: Would one of the options listed here make sense rather than using multiple splits? What you have is might be more efficiency, but a bit hard to read:
https://stackoverflow.com/questions/374326/how-to-get-filename-without-extension-from-file-path-in-ruby

app/services/tasks/sage_ingest_service.rb
```diff
@@ -0,0 +1,83 @@
...
+        unless file_names.count == 2
+          Rails.logger.tagged('Sage ingest') { Rails.logger.error("Unexpected package contents - more than two files extracted from #{package_path}") }
+          next
+        end
+        pdf_file_name = file_names.first
+        xml_file_name = file_names.last
+        # parse xml
+        # create object with xml and pdf
+        # save object
+        mark_pdf_done(unzipped_package_dir) if pdf_complete?(unzipped_package_dir, pdf_file_name)
```
bbpennel, line 30: Earlier in the loop you are checking that there are exactly 2 files in the package. Since it seems like both files must always be present, under what scenario will you not want to mark either the PDF or XML as done? Should the whole object be rolled back if either one is not present? Is this for a resumption scenario? I may be misunderstanding, but even in that scenario it seems like either both or neither files should be done.

app/services/tasks/sage_ingest_service.rb
```diff
@@ -0,0 +1,83 @@
...
+        mark_xml_done(unzipped_package_dir) if xml_complete?(unzipped_package_dir, xml_file_name)
+      end
+    end
```
bbpennel, line 33: Should there be a cleanup step for the extracted zip files after they are ingested? Are you anticipating needing to support resuming an incomplete batch of deposits, and if so, what mechanism are you hoping to use for that? I think we've used progress logs in some of the other scripts, which in this case would probably need to record the original file name. I wasn't sure if that was the intent of the .done files here, which would require keeping the extra unzipped files.

bbpennel (changes requested):
Looks good over all, a number of my questions probably stem from not being sure what the intent of the .done files is

app/services/tasks/sage_ingest_service.rb
maxkadel: - What I was thinking of for the .done file was trying to think through when we should delete or archive the files on the server, and that making sure there's a .done file could be a step for a separate cleanup step. Maybe this is pre-optimization? 
- We are going to need to do different types of processing for xml and pdf files - xml files get parsed and turned into the metadata for the object, pdf files get attached to the object and persisted (this is a hand-wave-y summary). Right now the x_complete? method is a placeholder for other checks we might want to do to make sure that everything is in its place for each type of document handling. 

So, that was my thinking, but I'm pondering whether it should be a single .done file and complete? method.

app/services/tasks/sage_ingest_service.rb
```diff
@@ -0,0 +1,83 @@
...
+
+    def mark_pdf_done(unzipped_package_dir)
+      FileUtils.touch(File.join(unzipped_package_dir, ".done.pdf"))
```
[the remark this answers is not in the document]
maxkadel: Good point - so, earlier in the process, check whether the .done file exists, and if so, move to the next object?

app/services/tasks/sage_ingest_service.rb
maxkadel: My thought was that we would want to clean up the working files after the entire batch was ingested - at the scale we have now, I don't anticipate that being a problem, but if it ramps up significantly we'll probably need to revisit it.

We do want to be able to resume incomplete batches, but I'm not totally clear on how we want to do that - I've tried to replicate the logging the proquest ingest has - is there another service you would recommend I look at?. 

Part of the intention of creating the .done file at this step is also just to make sure we're successfully unzipping the resources and know where the files are, in a way that a human can see at the end. We won't necessarily keep it, but it's scaffolding for the next steps.

spec/services/tasks/sage_ingest_service_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+    context "with a package already unzipped" do
+      before do
+        service.extract_files(first_zip_path)
+      end
+      it 'can write a .done file to the created directory' do
+        expect(Dir.exist?(first_dir_path)).to be true
+        expect(Dir.glob("#{first_dir_path}/*").count).to eq 2
+        expect do
+          service.mark_pdf_done(first_dir_path)
+        end.to change { Dir.entries(first_dir_path).count }.from(4).to(5)
```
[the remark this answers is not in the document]
maxkadel: Yup, good idea.

spec/services/tasks/sage_ingest_service_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+        expect do
+          service.mark_xml_done(first_dir_path)
+        end.to change { Dir.entries(first_dir_path).count }.from(5).to(6)
+      end
+    end
+
+
+    context "with a give file already present" do
```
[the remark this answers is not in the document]
maxkadel: typo - good catch

app/services/tasks/sage_ingest_service.rb
bbpennel: We will need to do different types of processing for the two files, but it seems like attaching the PDFs won't happen unless the XML parsing succeeds? And we wouldn't want to commit the work unless the PDF attached successfully.  I can't remember if you can wait until the work and its filesets have been created before saving everything. If so, then you would only need to track that the package completed. If not, then you would probably need to record both work creation and pdf attachment separately.

The zip file will still exist after extracting it, won't it? So cleaning up the extracted files should not pose any data loss risk. Only needing to record a single entry per package would be preferable, whether its a .done file or a progress log.

app/services/tasks/sage_ingest_service.rb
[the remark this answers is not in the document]
bbpennel: yes, having a check for if an object is done as early in the loop as possible would make sense.

app/services/tasks/sage_ingest_service.rb
bbpennel: Its really an implementation detail how progress gets tracked, so I don't want to push on this too much. Apparently the progress log setup isn't used for the proquest script, maybe because it is fairly small batches. The utility I was thinking of is this thing:
https://github.com/UNC-Libraries/hy-c/blob/431427d061465d3aa5ba2ac6e0687a2ad98b3a5b/lib/tasks/migrate/services/progress_tracker.rb

Its used by a number of the bulk utility scripts in hyc and hyc-utils:
https://github.com/UNC-Libraries/hy-c/search?q=ProgressTracker

It's up to you which approach you prefer and think will work best for this scenario.

app/services/tasks/sage_ingest_service.rb
maxkadel: Ok, that makes sense to me, I can change that, and adjust the ticket to reflect that.

app/services/tasks/sage_ingest_service.rb
```diff
@@ -0,0 +1,61 @@
...
+    end
+
+    def mark_done(orig_file_name)
+      Rails.logger.tagged('Sage ingest') { Rails.logger.info("Marked package ingest complete #{orig_file_name}") }
+      @ingest_progress_log.add_entry(orig_file_name)
+    end
+
+    # TODO: Make more assertions about what a completed ingest looks like and test here
+    def package_ingest_complete?(dir, file_names)
+      return true if File.exist?(File.join(dir, file_names.first)) && File.exist?(File.join(dir, file_names.last))
```
bbpennel, line 43: I'm not totally sure what this checks purpose is after the object has already been ingested, but I understand that we don't know what all the assertions are to verify that the package was ingested yet, so its fine for now

spec/services/tasks/sage_ingest_service_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+  context "with a package already unzipped" do
+    before do
+      # empty the progress log
+      File.open(ingest_progress_log_path, 'w') {|file| file.truncate(0) }
+    end
+    it 'can write to the progress log' do
+      allow(service).to receive(:package_ingest_complete?).and_return(true)
+      expect(File.size(ingest_progress_log_path)).to eq 0
+      service.mark_done(first_package_identifier)
+      expect(File.size(ingest_progress_log_path)).to eq 37
```
bbpennel, line 56: up to you if you want to change it, but the contents of the progress log should be pretty predictable, so you could assert that the progress log contains the expected filename rather than or in addition to checking the file size.

PR #735, opened 2021-12-17
Rubocop part 0

maxkadel:
Pared down Rubocop todo list - this change only incorporates Layout changes, which should be non-breaking.


app/controllers/concerns/hyc/download_analytics_behavior.rb
```diff
@@ -14,17 +14,17 @@ def track_download
...
-                                      category: @admin_set_name,
-                                      data_source: 'server-side',
-                                      hostname: request.host,
-                                      label: params[:id],
-                                      linkid: request.url,
-                                      referrer: request_referrer,
-                                      user_agent: request.headers['User-Agent'],
-                                      user_ip: request.remote_ip
-                                     )
+            action: 'DownloadIR',
```
bbpennel, line 17: I don't feel strongly about it, but I'm not totally sure about this correction, its a pretty common practice to indent parameters to the depth of the ( they correspond with

app/indexers/hyrax/deep_indexing_service.rb
```diff
@@ -90,12 +93,13 @@ def append_label_and_uri(solr_doc, solr_field_key, field_info, val)
       full_label = parse_geo_request(val.to_uri.to_s)
       val = val.solrize(full_label)
       create_and_insert_terms_handler.create_and_insert_terms(solr_field_key,
-                                         val.first,
-                                         field_info.behaviors, solr_doc)
+                                                              val.first,
```
bbpennel, line 96: I'm kind of confused why it wants these parameters indented to the ( but for the other file it removed that indent. Is it because this one has the first parameter on the same line as the opening (, while the other had the first param on its own line?

app/lib/mini_magick/image/info.rb
```diff
@@ -57,13 +57,13 @@ def cheap_info(value)
...
 
           @info.update(
-              "format"     => format,
-              "width"      => Integer(width),
-              "height"     => Integer(height),
-              "dimensions" => [Integer(width), Integer(height)],
-              "size"       => File.size(path),
-              "human_size" => size,
-              )
+            "format"     => format,
```
bbpennel, line 60: I feel a little iffy about making large numbers of formatting changes to files which are overrides from other gems, it may make it harder to pull in updates from upstream in the future. On the other hand, making exceptions for every overridden file also sounds awful

lib/tasks/migrate/services/metadata_parser.rb
```diff
@@ -266,9 +266,9 @@ def parse
           
           MigrationHelper.retry_operation('adding collection') do
             work_attributes['member_of_collections'] = Array(Collection.create(
-                title: [@collection_name],
-                depositor: @depositor.uid,
-                collection_type_gid: user_collection_type))
+                                                               title: [@collection_name],
```
bbpennel, line 269: Huh, I have no idea how it decided on this indent

app/indexers/hyrax/deep_indexing_service.rb
maxkadel: Yes, it enforces slightly different style depending on whether the opening parenthesis is on its own line or not.

app/controllers/concerns/hyc/download_analytics_behavior.rb
maxkadel: huh, yeah, usually that's what Rubocop does, not sure what happened with this one. I'll try to fix it.

maxkadel: I've manually gone through the examples of where it changed parentheses indentation for this pass. 

I started out making all of the clauses align with their parent parenthesis, but this caused issues in some cases because the arguments were so long, and the parenthesis started so far into the line, that the important info was falling off my fairly wide monitor. I've tried to make them more consistent and readable, where there are multiple parentheses that a chunk could belong to. 

I think as we move forward, we might want to look at these nested parentheses within parentheses and see if we can factor out some methods to make them more readable and clear.

bbpennel (approved):

PR #739, opened 2022-01-03
Allow authority to find http or https

maxkadel:
Connected to https://jira.lib.unc.edu/browse/HYC-1319

Connected to https://jira.lib.unc.edu/browse/HYC-1280

Sage provides the URIs for licenses with "https", which we only realized when I tried to find the label for them. 

I talked to Anna and Rebekah about being agnostic about mapping URIs to the same authority record whether they use "http" and "https" at the beginning, and they were fine with that. 

We will still only store / return records with "http" as the id, but "#find" will return the record for ids starting with either http or https.

This PR also adds a test for one of our overrides - Qa::Authorities::Local::FileBasedAuthority - this override does not appear to have been tested specifically previously.

### DIFF
--- a/spec/lib/qa/authorities/local/file_based_authority_spec.rb
+++ b/spec/lib/qa/authorities/local/file_based_authority_spec.rb
@@ -11,7 +11,6 @@ RSpec.describe Qa::Authorities::Local::FileBasedAuthority do
       expect(licenses.all).to be_instance_of(Array)
       expect(licenses.all.first).to be_instance_of(HashWithIndifferentAccess)
     end
-
     it "has the expected keys in those hashes" do
       expect(licenses.all.first.keys).to match_array(['id', 'label', 'active'])
     end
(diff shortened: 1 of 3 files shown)


spec/lib/qa/authorities/local/file_based_authority_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+
+    context "with no matching entries" do
+      let(:expected) { [] }
+      it "returns an empty array" do
+        expect(licenses.search("penguins")).to eq(expected)
+      end
+    end
+  end
+
+  describe "#find" do
```
bbpennel, line 59: Is the find method case insensitive like the search method? Could add a test to verify the correct behavior in whichever direction is correct.

bbpennel (approved):
This looks good, left a few questions/suggestions but nothing essential

PR #741, opened 2022-01-11
HYC-1333: Latest release of hydra-derivs fixes triple-pdf issue

maxkadel:
See https://github.com/samvera/hyrax/issues/4971
and https://github.com/samvera/hydra-derivatives/pull/227 

I'd like to deploy to cdr-test and test whether re-creating derivatives with this resolves the issue. 

Unknowns: 
- Do we need to re-create all derivatives for a work, or only thumbnails?
  - If only thumbnails, how do we do that? 
- How can we find works to target?
  - (Finding works where the attached file is a PDF is probably relatively easy, but limiting to monochrome first pages seems like it would be difficult.

image


maxkadel:
A. We're actually overriding the file that was updated
B. It doesn't fix the problem for the version of ImageMagick we're currently stuck on
C. Pushing up a PR that fixes it with GraphicsMagick

PR #746, opened 2022-01-18
HYC-1282: Attach PDFs to object ingested through Sage

maxkadel:
Deployed to cdr-test using admin set https://cdr-test.lib.unc.edu/admin/admin_sets/c534fq757 

- If the app can't map the license uri to the controlled vocabulary, log a warning to the Rails log (we still want the object to save / do the rest of the ingest, but we want to know about it so that we can add that license to the controlled vocabulary and add it to the object that's missing it)
- Update the sage logger to log to its own file in the logs directory, which is now configured with sane defaults 
- Raise errors if the admin set or user from the configuration can't be found (the job will fail eventually, might as well fail early) 
- Do not unzip the file within a temp directory block, because then the background jobs won't be able to find it
  - Will need to add these files to the cleanup cron job
- Create the file_set with the PDF and attach it to the work
- Refactor the Sage feature test not to use a before(:all) block - it's less good practice to use :all and it was making mocking things difficult. Found other ways to make test somewhat faster (combined some tests with shared setup)


bbpennel (approved):
This looks good. Left one question. Only other thing I was wondering, I notice from the proquest script that we are explicitly making sure it gets put into a sipity workflow. Not sure if that's needed here or not

app/services/tasks/sage_ingest_service.rb
```diff
@@ -15,29 +26,39 @@ def initialize(args)
...
-          xml_file_name = file_names.find { |name| name.match(/^(\S*).xml/) }
-          # parse xml
-          ingest_work = JatsIngestWork.new(xml_path: File.join(dir, xml_file_name))
-          # Create Article with metadata and save
-          build_article(ingest_work)
-          # Add PDF file to Article (including FileSets)
-          # save object
-          # set off background jobs for object?
-          mark_done(orig_file_name) if package_ingest_complete?(dir, file_names)
+        dir = Rails.root.join('tmp', 'sage_files', orig_file_name)
```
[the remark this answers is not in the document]
maxkadel: Yeah, it actually saves to the /net/deploy…/temp directory, which is no good because then if we deploy before all the background jobs finish, the files will be wiped out.

maxkadel:
> This looks good. Left one question. Only other thing I was wondering, I notice from the proquest script that we are explicitly making sure it gets put into a sipity workflow. Not sure if that's needed here or not

I’m not sure either - from what it looks like on cdr-test I’m not seeing anything missing, but I’ll look into it more. (I wrote this mostly tdd / bdd style)

spec/services/tasks/sage_ingest_service_spec.rb
```diff
@@ -83,14 +83,16 @@
         expect(service.ingest_progress_log).to be_instance_of(Migrate::Services::ProgressTracker)
       end
     end
-
+    # rubocop:disable Layout/MultilineMethodCallIndentation
```
bbpennel, line 86: seeing this makes me wonder if we should consider turning off this rule? the multiline call looks reasonable to me

bbpennel (approved):

spec/services/tasks/sage_ingest_service_spec.rb
maxkadel: I think it's helpful for everything except for this particular rspec thing (the `.and change`) - if I had infinite time I would write a PR to rubocop-rspec for it, but in the meantime I think wrapping these in a disable is worthwhile.

maxkadel:
wrt Sipity - I think that's only needed if you have a special workflow, or are creating the admin set on the command line, rather than via the UI.

PR #751, opened 2022-01-27
HYC-1282: Add Sage Articles to workflow

maxkadel:
- Factories borrowed from Hyrax factories of the same names
- Sipity makes me very unhappy

Deployed to cdr-test, rake task run on https://cdr-test.lib.unc.edu/catalog?f%5Badmin_set_sim%5D%5B%5D=Sage+test+five+admin+set&locale=en - they appear to have workflows!

### DIFF
--- a/spec/services/tasks/sage_ingest_service_spec.rb
+++ b/spec/services/tasks/sage_ingest_service_spec.rb
@@ -15,18 +15,32 @@ RSpec.describe Tasks::SageIngestService, :sage do
   let(:first_pdf_path) { "#{path_to_tmp}/10.1177_1073274820985792.pdf" }
   let(:first_xml_path) { "#{sage_fixture_path}/#{first_package_identifier}/10.1177_1073274820985792.xml" }
   let(:ingest_progress_log_path) { File.join(sage_fixture_path, 'ingest_progress.log') }
+
+  let(:permission_template) do
+    FactoryBot.create(:permission_template, source_id: admin_set.id)
+  end
+  let(:workflow) do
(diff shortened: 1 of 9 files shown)


spec/services/tasks/sage_ingest_service_spec.rb
```diff
@@ -15,18 +15,32 @@
...
+    admin_set
+    permission_template
+    workflow
+    workflow_state
     # return the FactoryBot admin user when searching for uid: admin from config
     allow(User).to receive(:find_by).with(uid: 'admin').and_return(admin)
-    # instantiate the sage ingest admin_set
-    admin_set
+    # return the FactoryBot admin_set when searching for admin set from config
+    allow(AdminSet).to receive(:where).with(title: 'Open_Access_Articles_and_Book_Chapters').and_return([admin_set])
```
bbpennel, line 43: I wonder if it is possible/appropriate for the factory to add this particular mocking step, since the title is being provided as one of the creation parameters?

bbpennel (approved):
That's a lot of factorying to get the admin_sets and permission_template working, but I guess it saves us from having to create an reload actual objects

spec/services/tasks/sage_ingest_service_spec.rb
maxkadel: The problem is that we're not clearing out the database between tests, so we could end up with an admin set from a previous run of the test, which would have the same title but a different id.

maxkadel: I agree that it's a lot of mocking, but I think the issue is in the Sipity code in Hyrax, which is way too intertwined with too many different things :/

PR #758, opened 2022-02-16
HYC-1378: Tune thumbnail job

maxkadel:
Avoid memory problems and slow performance for thumbnail regeneration.
* Search in batches
* Only perform the thumbnail part of the derivatives job, do not run the fulltext extract task that's bundled with it in Hyrax.


app/jobs/create_pdf_thumbnail_job.rb
```diff
@@ -0,0 +1,34 @@
...
+  # Queueing as import so that it doesn't block works imported via the UI
+  queue_as :import
+
+  def perform(file_set_id:, file_id:)
+    file_set = FileSet.find(file_set_id)
+
+    # There is no method I can find to directly ask "What is the mime_type of the file with ID foo?", so this is my
+    # indirect way of doing so
+    mime_type_map = file_set.files.map { |file| { mime_type: file.mime_type, file_id: file.id } }
+    target_file = mime_type_map.select { |file| file[:file_id] == file_id }&.first
```
bbpennel, line 15: Partially to make sure I understand this block, but could we select the target file directly, skip the mime_type_map, and then check its mimetype directly against that? So
```
target_file = file_set.files.select { |file| file.id == file_id }&.first
...
return unless target_file.mimetype == 'application/pdf'
```
Although it may not matter much since this is temporary code

bbpennel (approved):
Looks good, left one question but you're welcome to ignore it

app/jobs/create_pdf_thumbnail_job.rb
maxkadel: Dang it, I could have sworn I tried a million ways to get the mime_type directly from the target file yesterday, and couldn't get it to work (I would have sworn it was going to say there wasn't a method for mime_type for the file). So simple. So sensible.

maxkadel: oh wait, it's not consistent - I tried it on the command line, but then when I tested it:

  1) CreatePdfThumbnailJob saves the file to the derivatives directory
     Failure/Error: return unless target_file.mimetype == 'application/pdf'

     NoMethodError:
       undefined method `mimetype' for #<Hydra::PCDM::File:0x000000001209b728>
     # ./app/jobs/create_pdf_thumbnail_job.rb:15:in `perform'
     # /home/vagrant/.gem/ruby/gems/i18n-1.8.10/lib/i18n.rb:309:in `with_locale'
     # ./spec/jobs/create_pdf_thumbnail_job_spec.rb:30:in `block (2 levels) in <top (required)>'
     # /home/vagrant/.gem/ruby/gems/webmock-3.14.0/lib/webmock/rspec.rb:37:in `block (2 levels) in <top (required)>'

config/environments/production.rb
```diff
@@ -46,7 +46,7 @@
 
   # Use the lowest log level to ensure availability of diagnostic information
   # when problems arise.
-  config.log_level = :warn
+  config.log_level = :debug
```
bbpennel, line 49: Probably don't want this in production? Unless its getting overridden somewhere else

spec/jobs/create_pdf_thumbnail_job_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    text_temp_path = /test.txt/
+    expect(Hydra::Derivatives::PdfDerivatives).to receive(:create).with(pdf_temp_path, any_args)
+    expect(Hydra::Derivatives::PdfDerivatives).not_to receive(:create).with(text_temp_path, any_args)
+    described_class.perform_now(file_set_id: file_set_with_extracted_text.id)
+  end
+
+  describe 'with a pdf with unexpected info' do
+    let(:file_set) { FactoryBot.create(:file_set, :with_malformed_pdf) }
+
+    # This test passes either when Rails.configuration.eager_load = true, which makes many other tests fail;
```
bbpennel, line 47: Based on this comment, I'm trying to figure out if the test is expected to pass or not? Which of these conditions is currently true?

maxkadel: It's marked as pending, so it doesn't run, but if we ran it as the code is now it would fail.

PR #764, opened 2022-03-01
HYC-1200: Add Solr 8 configuration in separate directory in order to run Solr 6 and Solr 8 at once

maxkadel:


solr/new_config/spellings.txt
```diff
@@ -0,0 +1,2 @@
+pizza
```
bbpennel, line 1: the most searched for word in the CDR

solr/new_config/schema.xml
```diff
@@ -0,0 +1,367 @@
...
+  - for best index size and searching performance, set "index" to false
+    for all general text fields, use copyField to copy them to the
+    catchall "text" field, and use that for searching.
+  - For maximum indexing performance, use the StreamingUpdateSolrServer
+    java client.
+  - Remember to run the JVM in server mode, and use a higher logging level
+    that avoids logging every request
+-->
+
+<schema name="Hydra Demo Index" version="1.5">
```
bbpennel, line 48: Are we sticking with version 1.5 for the time being? I know 1.6 is the current version, but note there is a TODO down below next to 1.6.

bbpennel (approved):
This looks good, left one question. I take it there are some adjustments to code coverage in this pr as well?

solr/new_config/schema.xml
maxkadel: I don't know that 1.6 actually exists? I think it's saying there should be a 1.6 that makes the change it mentions, but I don't think it's actually been created. And the linked issue is from 2018. Good times.

Should we change it to `name="Hy-C Index" version="1.0"` or something? And then we could change the original schema.xml to `name="Hy-C Index" version="0.1"`?

bbpennel: My understanding is that the version here is what solr uses to determine how to interpret the schema file in terms of syntax, defaults for attributes, etc. Solr 7 is the first version that supports schema version 1.6 (on a side note, its surprisingly hard to find examples of schema.xml files in the solr documentation):
https://github.com/apache/solr/blob/main/solr/server/solr/configsets/_default/conf/managed-schema.xml

It may be that hyrax can't work with schema version 1.6 yet, although i'm not sure what issues would be, it seems like the main change is `useDocValuesAsStored defaults to true`. It may not be worth worrying about right now, I was just wondering since we had recently switched the version in boxc.

PR #767, opened 2022-03-09
Add parallel docx thumbnails job, riffs on pdf thumbnails

maxkadel:


spec/jobs/create_docx_thumbnail_job_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    described_class.perform_now(file_set_id: file_set_one.id)
+  end
+
+  it 'stops processing if the FileSet and file are not docx files' do
+    ActiveJob::Base.queue_adapter = :inline
+    expect(Hydra::Derivatives::DocumentDerivatives).not_to receive(:create)
+    described_class.perform_now(file_set_id: file_set_three.id)
+  end
+
+  it 'runs the job without errors' do
```
bbpennel, line 31: What is the difference between this test and the one at line 19, in terms of what is being tested?

spec/jobs/regenerate_all_docx_thumbnails_job_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  before do
+    # Clean out index so we only find our example files
+    ActiveFedora::Cleaner.clean!
+    Blacklight.default_index.connection.delete_by_query('*:*')
+    Blacklight.default_index.connection.commit
+    # Mock virus checker so it passes in CI
+    allow(Hydra::Works::VirusCheckerService).to receive(:file_has_virus?) { false }
+    # instantiate FileSets
+    file_set_one
```
bbpennel, line 18: what is the difference between instantiating the filesets this way and using `let!` in lines 5-8?

bbpennel (approved):

spec/jobs/create_docx_thumbnail_job_spec.rb
maxkadel: The expect(SomeClass).to receive(:some_method) syntax "catches" the method call, and it is not actually run. So in order to test that it doesn't raise an error when the actual method is run, I have to have it in a separate test without the expect...to receive line

spec/jobs/regenerate_all_docx_thumbnails_job_spec.rb
maxkadel: It's about when they're instantiated - they need to be instantiated after Fedora and Blacklight have been cleared out, and after the VirusChecker mock is created.

PR #772, opened 2022-03-23
HYC-1364: Docker and ci parallel

maxkadel:
Allows developers to run application and dependencies on Docker, rather than Vagrant. 

There are some improvements I'd like to make, but I think at this point it's sufficiently functional to be merge-able. Some follow-up work is documented here - https://jira.lib.unc.edu/browse/HYC-1414 

Deployed to cdr-test on March 25, at 2:56 pm, working as expected. See also https://gitlab.lib.unc.edu/cdr/vagrant-rails/-/merge_requests/128


maxkadel:
With mutagen:

`Finished in 0.11253 seconds (files took 10.74 seconds to load)` - file loading took 1/3rd as much time, equivalent to Vagrant.

docker-compose.yml
```diff
@@ -0,0 +1,131 @@
...
+      - redis
+      - solr6
+    environment:
+      CLAMD_TCP_HOST: clamav
+      DATABASE_URL: postgresql://db:5432
+      FEDORA_TEST_URL: http://fedora:8080/rest
+      FEDORA_DEV_URL: http://fedora:8080/rest
+      FITS_LOCATION: '/fits/fits-1.0.5/fits.sh'
+      HYRAX_DATABASE_PASSWORD: password
+      HYRAX_HOST: https://localhost:3000
```
bbpennel, line 21: should this be http?

docker-compose.yml
```diff
@@ -0,0 +1,131 @@
+version: '3.7'
+
+services:
+  web: &web
+    build:
+      context: .
+    image: mkadel/hy-c
```
bbpennel, line 7: is the image persisted to dockerhub or anything? wondering if it should be `unclib/hy-c` or something

maxkadel: It is persisted to Dockerhub - we don't have an organizational dockerhub right now, and Jason said he was comfortable with it living on my dockerhub right now, but yes, we should probably move it somewhere else. There should be a barrier to doing this. Basically either pull or build the image as described, then push it to a different repository and change this line.

bbpennel (approved):

PR #776, opened 2022-04-04
HYC-1400: ProQuest affils that are not standard

maxkadel:
- Affiliations that can't be mapped to a standardized department ID will be recorded as other_affiliation
- The fact that it could not be mapped will be included in the ProQuest logs, nested within the logs for a specific object, so it will be clear which object should be re-visited.


bbpennel (approved):
Looks good!

maxkadel:
> Looks good!

Whoah! I think this is a first! Thanks :D

PR #777, opened 2022-04-04
HYC-1398: Add short label to department service

maxkadel:
Note: I updated the departments controlled vocabulary on the console, assigning each `short_label` to the current id. 

The idea is to allow the short_label and id to drift from each other, *without* the ids themselves needing to drift.

```ruby
vocab = YAML.load_file('config/authorities/departments.yml')
vocab['terms'].map { |term| term['short_label'] = term['id'] }
File.open('config/authorities/departments.yml', "w") { |file| file.write(vocab.to_yaml(line_width: -1)) }
```

### DIFF
--- a/app/services/departments_service.rb
+++ b/app/services/departments_service.rb
@@ -9,19 +9,24 @@ module DepartmentsService
     end
   end
 
+  # The permanent identifier for the term, stored in Fedora. This identifier should not be changed.
   def self.identifier(term)
     authority.all.reject { |item| item['active'] == false }.select { |department| department['label'] == term }.first['id']
   rescue StandardError
     nil
   end
 
-  def self.label(id)
+  # The full term associated with the identifier. This is currently used in the display of People objects
+  def self.term(id)
     authority.find(id).fetch('term')
   rescue StandardError
     Rails.logger.warn "DepartmentsService: cannot find '#{id}'"
     nil
   end
 
+  # The short version of the term associated with the identifier. These short terms were initially populated with the same
+  # values as the identifiers, but unlike the identifiers, these values *can* be changed without negative effects.
+  # This values is indexed to solr for department faceting.
   def self.short_label(id)
     authority.find(id).fetch('short_label')
   rescue KeyError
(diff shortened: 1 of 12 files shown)


app/services/departments_service.rb
```diff
@@ -23,6 +23,13 @@ def self.label(id)
     nil
   end
 
+  def self.short_label(id)
```
bbpennel, line 26: It might be helpful to add descriptions of what `short_label` is, how it relates to `term` and `id`, and how they are expected to be updated, which should not be modified, etc. I'm not sure if there's a better place to do that than the department_service, but this seems like a reasonable place for someone to look to figure out how the vocab works

lib/active_fedora/rdf/indexing_service.rb
```diff
@@ -191,6 +187,21 @@ def build_person_display(field_key, people)
       displays.flatten
     end
 
+    def build_affiliations(person_affiliation, display_text)
+      affiliations = split_affiliations(person_affiliation)
```
bbpennel, line 191: I think i follow what's happening in here, but it would be helpful to state what the starting value of `person_affiliation` comes from, like is it the `term`, the `id`, etc. I can't remember which form is stored in the person record.

bbpennel (commented):
I think this looks good, just left two suggestions for documentation

lib/active_fedora/rdf/indexing_service.rb
bbpennel: Now seems like a reasonable time to get the terminology lined up, we're not likely to circle back to it unless some other problem comes up

app/services/departments_service.rb
maxkadel: I've made `term` and the method to get the term match, and added documentation to each of these.

lib/active_fedora/rdf/indexing_service.rb
maxkadel: Cool, I've made a stab at it, let me know if it's not clearer.

bbpennel: Yup, that's much easier to understand now

bbpennel (approved):

PR #778, opened 2022-04-05
HYC-1359: List file sets without attached files

maxkadel:


app/services/file_set_remediation_service.rb
```diff
@@ -0,0 +1,31 @@
+# This class is to remediate FileSets where the file was not attached because the anti-virus service
```
bbpennel, line 1: Is this service going to be used for performing the actual remediation, once we figure out what that looks like? At the moment it seems to only be focused on searching for filesets without files.

app/services/file_set_remediation_service.rb
```diff
@@ -0,0 +1,31 @@
...
+          file_set = FileSet.find(solr_doc['id'])
+          next if has_files?(file_set)
+
+          csv << [file_set.id, Rails.application.routes.url_helpers.url_for(file_set)]
+        end
+      end
+    end
+  end
+
+  # rubocop:disable Naming/PredicateName
```
bbpennel, line 18: what does rubocop dislike about this name?

bbpennel (approved):
Looks like it should work as advertised. Does this need a rake task or anything like that for executing it?

I also appreciate that clam caused this pr to fail tests the first time around, very fitting.

app/services/file_set_remediation_service.rb
maxkadel: Ah, yes, it does need a rake task and background job. Good times!

I'm thinking we can add the actual remediation code to this class, once we decide what to do. I think the hyc_crawler_service kinda went off the rails... I blame the `yield` I borrowed from Hyrax. </half-joke, half-serious>

app/services/file_set_remediation_service.rb
maxkadel: That it starts with a predicate? It suggests just "files", but I find that really confusing to read.

maxkadel: https://docs.rubocop.org/rubocop/cops_naming.html

bbpennel: as a java developer, those "bad" examples hurt

app/services/file_set_remediation_service.rb
```diff
@@ -7,9 +7,14 @@ def create_csv_of_file_sets_without_files
       FileSet.search_in_batches('*:*') do |group|
         group.each do |solr_doc|
           file_set = FileSet.find(solr_doc['id'])
+          next unless file_set
```
bbpennel: are both this next and the ObjectNotFoundError rescue necessary? I'm just trying to figure out what gets skipped here that doesn't trigger the rescue

maxkadel: Oh, somehow I didn't realize you could put a `next` in a rescue block, but I see now you can. But now it's in its own method, so that doesn't work anymore, and when I smoosh the methods back together with the 'next' in the rescue, CodeClimate yells about complexity (and I'm inclined to agree). But I've learned something new!

PR #780, opened 2022-04-07
Add a version service

maxkadel:
- Makes version information more testable
- Avoids errors using mutagen with Docker

### DIFF
--- a/spec/services/version_service_spec.rb
+++ b/spec/services/version_service_spec.rb
@@ -38,10 +38,8 @@ RSpec.describe VersionService do
 
   context 'in a deployed environment' do
     before do
-      Rails.logger.debug("in test 'before' block, before 'allow': #{Rails.env}")
       allow(Rails).to receive(:env).and_return(ActiveSupport::StringInquirer.new('production'))
-      allow(described_class).to receive(:directory_name).and_return('/net/deploy/ir/test/releases/20211202095413')
-      Rails.logger.debug("in test 'before' block, after 'allow': #{Rails.env}")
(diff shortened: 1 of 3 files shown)


spec/services/version_service_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+    it 'returns a branch name' do
+      expect(described_class.branch).to eq('my-cool-tag')
+    end
+  end
+
+  context 'in a deployed environment' do
+    before do
+      Rails.logger.debug("in test 'before' block, before 'allow': #{Rails.env}")
+      allow(Rails).to receive(:env).and_return(ActiveSupport::StringInquirer.new('production'))
+      allow(described_class).to receive(:directory_name).and_return('/net/deploy/ir/test/releases/20211202095413')
```
bbpennel, line 43: I might prefer to use a more fake path, like 
```suggestion
      allow(described_class).to receive(:directory_name).and_return('/path/to/releases/20211202095413')
```

spec/services/version_service_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+    it 'can give a deploy date' do
+      expect(described_class.deploy_date).to eq('20211202095413')
+    end
+
+    it 'can give a phrase about the deploy date' do
+      expect(described_class.deploy_date_phrase).to eq('Deployed on December 2, 2021')
+    end
+    context 'with an unexpected directory name' do
+      before do
+        allow(described_class).to receive(:directory_name).and_return('/net/deploy/ir/test/releases/cheeseburger')
```
bbpennel, line 56: 🍔 (same suggestion about faker path) 🍔

PR #782, opened 2022-04-14
HYC-1366: Add mappings from ProQuest to departmental vocab

maxkadel:


bbpennel (approved):
I'm good with these changes as long as Anna or Rebekah are

PR #783, opened 2022-04-18
Tests random order

maxkadel:


spec/features/create_work_in_honors_thesis_workflow_spec.rb
```diff
@@ -7,44 +7,44 @@
 RSpec.feature 'Create and review a work in the honors thesis workflow', js: false do
   context 'a logged in user' do
     let(:user) do
-      User.new(email: 'test@example.com', guest: false, uid: 'test') { |u| u.save!(validate: false) }
+      FactoryBot.create(:user, email: 'test@example.com', guest: false, uid: 'test')
```
bbpennel, line 10: Just curious, why did the users in this test need custom parameters rather than a randomized UID? I'm not seeing references to the uid or the email in the test

bbpennel (approved):
Good work, left one question

spec/features/create_work_in_honors_thesis_workflow_spec.rb
maxkadel: I don't think it absolutely needs them, but I think it can make debugging easier

PR #785, opened 2022-04-21
HYC-1273: Remove one science

maxkadel:
We are not longer using the OneScience ingests


bbpennel (changes requested):
There are a bunch of onescience classes called out in the rubocop todo list, might as well clear those out at the same time as purging the rest of it
https://github.com/UNC-Libraries/hy-c/blob/remove_one_science/.rubocop_todo.yml#L90

bbpennel (approved):

PR #786, opened 2022-04-22
HYC-1273: Remove deprecated migration code

maxkadel:


bbpennel (changes requested):
Looks like we can also clean up these methods
https://github.com/UNC-Libraries/hy-c/blob/remove_deprecated_code/lib/tasks/migration_helper.rb#L2-L40
and the tests for them
https://github.com/UNC-Libraries/hy-c/blob/remove_deprecated_code/spec/helpers/migration_helper_spec.rb#L5-L71

otherwise, everything looks good to me

bbpennel (approved):

PR #799, opened 2022-06-03
HYC-1446: Update Docker configs to ruby 2.7

cazzerson:


bbpennel (approved):
Works for me locally, and deploying to servers after updating hyc-gems

PR #800, opened 2022-06-03
HYC-1445: update dependencies to latest minor version

cazzerson:


bbpennel:
Included in https://github.com/UNC-Libraries/hy-c/pull/804

PR #803, opened 2022-06-15
Update to passenger 6

lfarrell:
https://jira.lib.unc.edu/browse/HYC-1454


lfarrell:
Changing build environments in the middle of a check apparently wasn't a good idea as pr can't be merged.

PR #805, opened 2022-06-20
Fix deprecation warnings

lfarrell:
https://jira.lib.unc.edu/browse/HYC-1448


bbpennel (approved):
This resolves the worst of the DEPRECATION spam, that's for sure. There are still lots of other warnings of course, which would be nice to address (the ticket mentions a few on the advanced search page that are still happening). We could probably merge this as is and address more warnings in other PRs if needed, although i recognize its probably a losing battle, and maybe pointless before we migration to hyrax 3

PR #810, opened 2022-06-22
HYC-1460 - Solr migration task

bbpennel:
https://jira.lib.unc.edu/browse/HYC-1460

* Adds rake tasks for indexing all objects in one solr instance into a second solr instance.
    * Supports resumption and more info about progress
    * Supports performing followup partial indexing requests
* Adds environment variable to allow a different local_env.yml file be specified (must be in the conf directory)

### DIFF
--- a/.codeclimate.json
+++ b/.codeclimate.json
@@ -22,7 +22,7 @@
     "method-complexity": {
       "enabled": true,
       "config": {
-        "threshold": 5
+        "threshold": 24
       }
     },
     "method-count": {
@@ -34,7 +34,7 @@
     "method-lines": {
       "enabled": true,
       "config": {
-        "threshold": 25
+        "threshold": 130
       }
     },
     "nested-control-flow": {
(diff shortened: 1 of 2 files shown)


lib/tasks/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+There must be a separate local_env.yml in the conf folder with the configuration for the second solr instance, which will be selected using the LOCAL_ENV_FILE environment variable.
+
+Usage:
+
+1. Generate listing of ids of currently indexed objects
+```
+$ LOCAL_ENV_FILE=local_env.yml bundle exec rake migrate_solr:list_ids -- -o /tmp
+
+Stored id list to file: /tmp/id_list_2022-06-21T19_55_08Z.txt
```
sharonluong, line 13: how does one get this timestamp? does the command use this specific one?

bbpennel: This is an example of the output from the command. The timestamp in the filename is the the current time when the script begins to execute (minus a minute). For the `-a` option, you can give it any (iso8601) timestamp you want, but the intent is for users to give it the timestamp from the previous id list run, so that you would get everything updated after that run.

sharonluong (approved):

PR #817, opened 2022-06-30
HYC-1475 - Index contributor label field

bbpennel:
https://jira.lib.unc.edu/browse/HYC-1475

* Index contributor_label_tesim, which is needed for the advanced search page

### DIFF
--- a/lib/active_fedora/rdf/indexing_service.rb
+++ b/lib/active_fedora/rdf/indexing_service.rb
@@ -29,6 +29,7 @@ module ActiveFedora::RDF
       @person_label = []
       @creator_label = []
       @advisor_label = []
+      @contributor_label = []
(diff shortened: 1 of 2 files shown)


lib/active_fedora/rdf/indexing_service.rb
```diff
@@ -171,6 +176,8 @@ def build_person_display(field_key, people)
             @creator_label.push(Array(person['name']))
           elsif field_key.to_s == 'advisors'
             @advisor_label.push(Array(person['name']))
+          elsif field_key.to_s == 'contributors'
+            @contributor_label.push(Array(person['name']))
```
sharonluong, line 180: does `Array(person['name'])` change between elsifs? i see it repeated a lot.

PR #822, opened 2022-07-08
Fix issue with horizontal scroll bar showing up regardless of page width

lfarrell:


bbpennel (approved):
Horizontal scroll begone

PR #824, opened 2022-07-12
Check that kwd-group isn't nil for Sage ingests

lfarrell:
https://jira.lib.unc.edu/browse/HYC-1471


lfarrell:
https://jira.lib.unc.edu/browse/HYC-1471

bbpennel (approved):

PR #828, opened 2022-07-18
HYC-1493 - Coverage/overrides for statistics classes

bbpennel:
https://jira.lib.unc.edu/browse/HYC-1493

* Adds test coverage for overrides in statistic.rb, file_download_stats.rb, and download.rb
* Converts statistic.rb and file_download_stats.rb to overrides (not sure how to do/verify this for download.rb so left it alone)

PR #829, opened 2022-07-18
HYC-1514 - Cache redirect mappings

bbpennel:
https://jira.lib.unc.edu/browse/HYC-1514

* Caches redirect_uuids.csv into two hashes, one for lookups by uuid and the other for lookups by new path/id
* Adds more testing of redirect lookups

This will need to be rebased after https://github.com/UNC-Libraries/hy-c/pull/828 is merged


spec/helpers/hyc_helper_spec.rb
```diff
@@ -55,14 +55,42 @@
...
+    context 'with multiple mappings' do
+      before do
+        File.open(ENV['REDIRECT_FILE_PATH'], 'w') do |f|
+          f.puts 'uuid,new_path'
+          f.puts "02fc897a-12b6-4b81-91e4-b5e29cb683a6,articles/#{article.id}"
+          f.puts "b19ac193-3fc1-43b5-ac52-b4828ec0afdb,parent/#{article.id}/file_sets/ff365f05d"
+        end
+      end
+
+      it 'returns redirect mapping' do
```
sharonluong, line 86: what about a test that looks for a uuid that does not exist in the redirect mappings file?

spec/features/boxc_redirect_spec.rb
```diff
@@ -30,6 +31,7 @@
     ENV['REDIRECT_FILE_PATH'] = cached_redirect_file_path
     ENV['REDIRECT_OLD_DOMAIN'] = cached_redirect_old_domain
     ENV['REDIRECT_NEW_DOMAIN'] = cached_redirect_new_domain
+    HycHelper.clear_redirect_mapping
```
sharonluong, line 34: why does this have to be put at the beginning and the end?

bbpennel: i was finding that when running all the tests there would sometimes be an empty mapping cached from another test that would impact these tests, depending on the random order they ran in. My guess is that another test was making a call to the redirect mapping, which loaded the default empty mapping file.

spec/helpers/hyc_helper_spec.rb
bbpennel: it should be the same as https://github.com/UNC-Libraries/hy-c/blob/2361b85a0cba8f4ab5c757646cac877d0ceb87f2/spec/helpers/hyc_helper_spec.rb#L67 but i've added in another expectation with a uuid.

spec/features/boxc_redirect_spec.rb
sharonluong: It feels kinda weird that tests affect others so much, is it because of the class variables?

bbpennel: yes, it seemed like the best option so that it wouldn't end up with the mapping loaded/cached multiple times (HycHelper is included in 9 classes, although only 3 of them reference redirect_lookup). I'm open to making it a instance variable though if that seems preferable. Having a single instance of a service which is responsible for redirects seems better to me than including a helper, but i'm not really clear on how it would end up in all the right places (my lack of rails knowledge showing)

sharonluong: Hmm. i guess ideally this redirect stuff would be in a service class, not a helper. Then you could initialize the class with the instance variables for the mapping. I'm not familiar with where this is called so I'm not sure if that's persistent enough for your needs.  Maybe would you be able to use the rails cache?

bbpennel: I don't want this ticket to expand too much, but i can see if it looks like a reasonable refactoring this afternoon

sharonluong (approved):
Looks great, thanks!

PR #836, opened 2022-07-26
Change sync mode for mount-code

bbpennel:
* Switch mount-code to using two-way-resolved sync mode, to prevent conflicts from stopping syncs of updated files in git from reaching the container

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -108,6 +108,7 @@ volumes:
 x-mutagen:
   sync:
     defaults:
+      mode: "two-way-resolved"
       ignore:
         vcs: true
         paths:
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -110,9 +110,12 @@ x-mutagen:
     defaults:
       ignore:
         vcs: true
+        paths:
+          - ".DS_Store"
     mount-code:
       alpha: '.'
       beta: "volume://mount-code"
+      mode: "two-way-resolved"
```
cazzerson, line 118: I wonder if we should change this for the solr volumes as well? Or set it as a default?

bbpennel: We could probably try doing this for all volumes, i did actually have a conflict in my solr8-data mount with the schema.xml, which could probably cause some weirdness, but has not so far to the best of my knowledge.

cazzerson: Seems like it would have the same issues as it's part of the code directory. Maybe just move `mode` to the `sync:defaults:` section?

cazzerson (approved):

PR #839, opened 2022-08-11
Sandi overrides

bbpennel:
* Adjustments to the way overrides work for two statistics classes to reduce the amount of upstream code we are duplicating. Improvement, or not?

### DIFF
--- a/app/overrides/models/hyrax/statistic_override.rb
+++ b/app/overrides/models/hyrax/statistic_override.rb
@@ -26,24 +26,16 @@ Hyrax::Statistic.class_eval do
     private
 
     # [hyc-override] add error handling for timeouts
+    alias :original_combined_stats :combined_stats
     def combined_stats(object, start_date, object_method, ga_key, user_id = nil)
-      stat_cache_info = cached_stats(object, start_date, object_method)
-      stats = stat_cache_info[:cached_stats]
-
-      if stat_cache_info[:ga_start_date] < Time.zone.today
-        begin
-          ga_stats = ga_statistics(stat_cache_info[:ga_start_date], object)
-          ga_stats.each do |stat|
-            lstat = build_for(object, :date => stat[:date], object_method => stat[ga_key], :user_id => user_id)
-            lstat.save unless Date.parse(stat[:date]) == Time.zone.today
-            stats << lstat
-          end
-        rescue Net::ReadTimeout
-          Rails.logger.warn "Unable to retrieve GA stats for #{object.id}. Request timed out. Using cached stats for object."
-        end
+      begin
+        original_combined_stats(object, start_date, object_method, ga_key, user_id)
+      rescue Net::ReadTimeout
+        Rails.logger.warn "Unable to retrieve GA stats for #{object.id}. Request timed out. Using cached stats for object."
+        # return the cached statistics since we can't get fresh ones
+        stat_cache_info = cached_stats(object, start_date, object_method)
+        stats = stat_cache_info[:cached_stats]
       end
-
-      stats
     end
   end
 end
--- a/app/overrides/models/hyrax/statistic_override.rb
+++ b/app/overrides/models/hyrax/statistic_override.rb
@@ -26,7 +26,7 @@ Hyrax::Statistic.class_eval do
     private
 
     # [hyc-override] add error handling for timeouts
-    alias :original_combined_stats :combined_stats
+    alias_method :original_combined_stats, :combined_stats
 
     def combined_stats(object, start_date, object_method, ga_key, user_id = nil)
       begin
--- a/app/overrides/models/hyrax/statistic_override.rb
+++ b/app/overrides/models/hyrax/statistic_override.rb
@@ -27,6 +27,7 @@ Hyrax::Statistic.class_eval do
 
     # [hyc-override] add error handling for timeouts
     alias :original_combined_stats :combined_stats
+
     def combined_stats(object, start_date, object_method, ga_key, user_id = nil)
       begin
         original_combined_stats(object, start_date, object_method, ga_key, user_id)
--- a/spec/models/file_download_stat_spec.rb
+++ b/spec/models/file_download_stat_spec.rb
@@ -49,7 +49,8 @@ RSpec.describe FileDownloadStat, type: :model do
         allow(described_class).to receive(:original_ga_statistics)
       end
 
-      it 'calls wrapper method succeed' do
+      it 'calls wrapped method with provided parameters' do
+        expect(described_class).to receive(:original_ga_statistics).with(start_date, file_set2)
         described_class.ga_statistics(start_date, file_set2)
       end
     end
--- a/spec/models/file_download_stat_spec.rb
+++ b/spec/models/file_download_stat_spec.rb
@@ -3,66 +3,55 @@ require 'rails_helper'
 require Rails.root.join('app/overrides/models/file_download_stat_override.rb')
 
 RSpec.describe FileDownloadStat, type: :model do
-  describe '.ga_statistic' do
-    context 'with fileset' do
-      let(:user) { FactoryBot.create(:user) }
-      let(:work) { Article.create(title: ['New Article']) }
-      let(:file_set) { FileSet.new }
-      let(:file_set_actor) { Hyrax::Actors::FileSetActor.new(file_set, user) }
+  let(:file_set) { FactoryBot.create(:file_set) }
+
+  context 'with redirect mappings' do
+    let(:tempfile) { Tempfile.new('redirect_uuids.csv', 'spec/fixtures/') }
+    before do
+      File.open(ENV['REDIRECT_FILE_PATH'], 'w') do |f|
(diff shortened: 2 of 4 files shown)


app/overrides/models/hyrax/statistic_override.rb
```diff
@@ -26,24 +26,16 @@ def ga_statistics(start_date, object)
...
-          end
-        rescue Net::ReadTimeout
-          Rails.logger.warn "Unable to retrieve GA stats for #{object.id}. Request timed out. Using cached stats for object."
-        end
+      begin
+        original_combined_stats(object, start_date, object_method, ga_key, user_id)
+      rescue Net::ReadTimeout
+        Rails.logger.warn "Unable to retrieve GA stats for #{object.id}. Request timed out. Using cached stats for object."
+        # return the cached statistics since we can't get fresh ones
+        stat_cache_info = cached_stats(object, start_date, object_method)
```
cazzerson, line 36: Is this in scope here?

cazzerson:
This is cool. Should the tests be updated as well? Seems like they could be much simpler now that we don't need to test Hyrax. Would it be reasonable to just test `as_subject()` and the rescue block in `combined_stats()`?

cazzerson:
To go full Sandi...I wonder if injecting the original methods would make testing even easier.

bbpennel:
I'm not quite sure what injecting the original methods would mean.

Part of me kind of likes that we're verifying that the output from our custom code plays nicely with the hyrax code, at least in so far as the test can tell. But yeah, can't probably pare down the tests now (might make code climate angry though!)

cazzerson:
Right, but I think the biggest payoff for this approach is simplifying our tests. It seems like we get several free compatibility checks with this approach. For example: 

1) If the aliased method is removed upstream, the alias operation should fail loudly.
2) If the number of required parameters in the upstream aliased method changes, calls to our method and our call to the original method should fail loudly.
3) If the class is removed upstream, `class_eval` should be missing (maybe this is a reason to stick with the `class_eval` approach?).
4) If other code starts calling the aliased upstream method with a different number of (optional) parameters, the call to our wrapper method should fail loudly.
5) If `ga_statistics()` started expecting something other than `id()` from the file parameter, the code should fail loudly.

This isn't every failure case, of course. Hyrax's `combined_stats()` could start catching exceptions and handling them differently than we'd like to. `ga_statistics()` could start modifying the file id. They could stop using these methods or the class but leave them in the codebase (ugh). So I guess my question is: how thoroughly should we test for these errors that could slip through the cracks? 

Re: injecting the original method, I was thinking about possibly accepting the method symbol as an optional argument to the wrapper methods. Reading a bit more about this, it sounds possible, but tricky, in Ruby. Another idea would be passing the `Hyrax::FileDownloadStat` instance in as an optional parameter to the wrapper methods, with `self` as the default value. Something like: `def ga_statistics(start_date, file, my_instance: self)` and then calling `my_instance.original_ga_statistics(start_date, file)`. Seems a little weird and tricky, but could let you pass in a fake for the upstream object.

spec/models/file_download_stat_spec.rb
```diff
@@ -3,66 +3,55 @@
...
+        allow(described_class).to receive(:original_ga_statistics)
+      end
 
-        it 'calls the Legato method with the correct path' do
-          expect(views).to receive(:for_file).with("#{file_set.id}|bfe93126-849a-43a5-b9d9-391e18ffacc6")
-          described_class.ga_statistics(start_date, file_set)
-        end
+      it 'calls wrapped method when override method is called' do
+        described_class.ga_statistics(start_date, file_set2)
+        expect(described_class.original_ga_statistics(start_date, file_set2))
```
cazzerson, line 54: should this be using `to receive().with()` syntax?

spec/models/hyrax/statistic_spec.rb
```diff
@@ -77,95 +77,37 @@ def filter(file)
...
       before do
-        allow(Hyrax::Analytics).to receive(:profile).and_raise(Net::ReadTimeout)
+        allow(concrete_stat_class).to receive(:original_combined_stats).and_raise(Net::ReadTimeout)
+        allow(concrete_stat_class).to receive(:cached_stats).and_return({cached_stats: :cached_results})
       end
 
-      it 'logs and handles error' do
-        expect(Rails.logger).to receive(:warn).with("Unable to retrieve GA stats for #{work_id}. Request timed out. Using cached stats for object.")
-        test_class.statistics(work, Time.zone.today - 4.days, user_id)
+      it 'returns stats from the original method' do
```
cazzerson, line 106: Is this an accurate description? Or is this just testing if it returns cached stats?

spec/models/hyrax/statistic_spec.rb
```diff
@@ -77,95 +77,37 @@ def filter(file)
     end
   end
 
-  describe '#statistics' do
-    let(:test_class) { WorkViewStat }
+  describe '.combined_stats' do
     let(:user) { FactoryBot.create(:user) }
     let(:user_id) { user.id }
```
cazzerson, line 82: do we still need these, or would simple fakes do?

spec/models/file_download_stat_spec.rb
```diff
@@ -3,66 +3,55 @@
...
-      let(:user) { FactoryBot.create(:user) }
-      let(:work) { Article.create(title: ['New Article']) }
-      let(:file_set) { FileSet.new }
-      let(:file_set_actor) { Hyrax::Actors::FileSetActor.new(file_set, user) }
-      let(:start_date) { 2.days.ago }
-
-      before do
-        file_set_actor.attach_to_work(work)
-        allow(Hyrax::Analytics).to receive(:profile).and_return(profile)
+  let(:file_set) { FactoryBot.create(:file_set) }
```
cazzerson, line 6: I'm tempted to cut most of this since `BoxcToHycRedirectService.redirect_lookup` should be tested elsewhere.

spec/models/hyrax/statistic_spec.rb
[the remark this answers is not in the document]
bbpennel: There's no check for an empty cache in the original method, so I was going with the same assumptions they represented
https://github.com/samvera/hyrax/blob/main/app/models/hyrax/statistic.rb#L80-L81

bbpennel: nope, I forgot to change it

spec/models/file_download_stat_spec.rb
bbpennel: There is a specific test for BoxcToHycRedirectService.redirect_lookup, which does verify cases like this. I'm tentative to remove usage of the real class here though. I usually prefer to use real methods from our own code base, particularly if there is only a single implementation, since its more realistic and we're less likely to run into issues where original class's behavior gets updated but we forget to adjust its usages, and the tests in the other class are still passing since their faking the old behavior.

spec/models/file_download_stat_spec.rb
bbpennel: Yes I believe so, will add it

spec/models/file_download_stat_spec.rb
cazzerson: Okay, that makes sense.

spec/models/file_download_stat_spec.rb
bbpennel: After spending quite a bit of time looking at this, I could not get the expect statement to work. I verified that the call to the `original_ga_statistics` method was happening in a few ways (code coverage, debug statement, having it return something and asserting on that, and commenting out the allow statement to see that it fails because it calls the original method), but for some reason the expectation would never find any calls to the method. Not sure if there's some special behavior needed for class methods, but my googling was not turning up anything relevant.

As far as I can tell, the following should work:
`expect(described_class).to receive(:original_ga_statistics)`
but produced
```
(FileDownloadStat(id: integer, date: datetime, downloads: integer, file_id: string, created_at: datetime, updated_at: datetime, user_id: integer) (class)).original_ga_statistics(*(any args))
    expected: 1 time with any arguments
    received: 0 times with any arguments
```
For the moment I have no assertion here. All I can think of to verify its being called is to have `original_ga_statistics` return or raise something.

spec/models/hyrax/statistic_spec.rb
bbpennel: switch it over to using a constant

spec/models/file_download_stat_spec.rb
cazzerson: I'm a bit out of my depth here, but I wonder if this is watching for a call to the class method (via `described_class`) when we should be using an instance double? One interesting discussion:

https://github.com/rspec/rspec-mocks/issues/1187#issuecomment-345560439

bbpennel: Ugh, I got through writing up a lengthy stackoverflow question, then while proofreading it I remembered that `expect` statements are supposed to go before the call to the method being tested. Now the test passes with the expectation.

PR #842, opened 2022-08-23
Only pull in the parts of the collection form we override and add tests

lfarrell:
https://jira.lib.unc.edu/browse/HYC-1488


spec/forms/hyrax/collection_form_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+                         :visibility,
+                         :collection_type_gid]
+    end
+  end
+
+  let(:collection) { Collection.new }
+  let(:repository) { double }
+  let(:form) { described_class.new(collection, nil, repository) }
+
+  describe '#secondary_terms' do
```
bbpennel, line 36: How do we decide whether to use a . or a # in front of the name of the method being tested? Is it based on whether it is a class or instance method, or something else? Wondering since `terms` is a class variable and `seconary_terms` is an instance method

spec/forms/hyrax/collection_form_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+                         :identifier,
+                         :related_url,
+                         :visibility,
+                         :collection_type_gid]
+    end
+  end
+
+  let(:collection) { Collection.new }
+  let(:repository) { double }
+  let(:form) { described_class.new(collection, nil, repository) }
```
bbpennel, line 34: If these are only used for `secondary_terms`, could probably move these inside its `describe`

spec/forms/hyrax/collection_form_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+                          :subject,
+                          :language,
+                          :identifier,
+                          :related_url,
+                          :resource_type
+                        ]
+    end
+  end
+
+  describe '.build_permitted_params' do
```
bbpennel, line 56: I'm not sure we need to test this, since we aren't overriding anything in it for collection_form. I assume its coming from one of the includes, and is based on the fields in terms, so should adapt as long as terms has the right value

bbpennel (changes requested):
Very satisfying seeing so much deletion!

spec/forms/hyrax/collection_form_spec.rb
lfarrell: Seems like . for class methods and # for instance methods, https://www.betterspecs.org/

bbpennel (approved):

PR #843, opened 2022-08-24
HYC-1490 - Test coverage/refactor for attach_files_to_work_job overrides

bbpennel:
https://jira.lib.unc.edu/browse/HYC-1490

* Add tests for overrides
* Change approach to override, removing non-overridden code, and wrapping the original method
* Change infected file behavior, so it now will only delete the pattern dir containing the file if its empty. Also, catch any errors that occur during deletion so email always gets sent.

### DIFF
--- a/spec/jobs/attach_files_to_work_job_spec.rb
+++ b/spec/jobs/attach_files_to_work_job_spec.rb
@@ -1,0 +1,52 @@
+require 'rails_helper'
+
+RSpec.describe AttachFilesToWorkJob, type: :job do
+  let(:user) { FactoryBot.create(:user) }
+  let(:admin) { FactoryBot.create(:admin) }
+  let(:role) { Role.new(name: 'admin') }
+  let(:work) { 
+    Article.create(title: ['New Article'],
+      depositor: user.user_key) 
+  }
+  let(:file_set) { FactoryBot.create(:file_set, title: ['test file set']) }
+  let(:file_set_actor) { Hyrax::Actors::FileSetActor.new(file_set, user) }
+  let(:admin_set) { AdminSet.create(title: ['an admin set']) }
+  let(:permission_template) { Hyrax::PermissionTemplate.create(source_id: admin_set.id) }
+  let(:workflow) { Sipity::Workflow.create(name: 'a workflow', permission_template_id: permission_template.id, active: true) }
+  let(:entity) { Sipity::Entity.create(workflow_id: workflow.id, proxy_for_global_id: work.to_global_id.to_s) }
+  let(:target_dir) { Dir.mktmpdir }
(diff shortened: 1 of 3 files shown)


spec/jobs/attach_files_to_work_job_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+      expect(File).not_to exist(target_file)
+    end
+
+    context 'that cannot be deleted' do
+      before do
+        allow(File).to receive(:delete).and_raise(Errno::EACCES)
+      end
+
+      it 'still sends virus email message' do
+        subject.perform(work, [uploaded_file])
```
lfarrell, line 49: There's no expect, just the action that needs to be checked.

spec/jobs/attach_files_to_work_job_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+      end
+    end
+  end
+
+  context 'with safe file' do
```
lfarrell, line 54: Shouldn't this end with an expect as well?

spec/jobs/attach_files_to_work_job_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+      # Original method should not be called
+      allow(subject).to receive(:original_perform).and_raise('nope')
+      allow(Hyc::VirusScanner).to receive(:hyc_infected?).and_return(ClamAV::VirusResponse.new(target_file, 'avirus'))
+      expect(Hyrax::Workflow::VirusFoundNotification).to receive(:send_notification)
+          .with(entity: entity, recipients: anything, comment: anything, user: user)
+    end
+
+    it 'sends virus email message' do
+      subject.perform(work, [uploaded_file])
+      expect(File).not_to exist(target_file)
```
lfarrell, line 40: Seems like this isn't testing whether an email was sent.

lfarrell (changes requested):
Added a few comments, though I may just be misunderstanding what the tests are doing.

spec/jobs/attach_files_to_work_job_spec.rb
bbpennel: the expect is up in the `before` section. I could move it into the `it`, but would have to duplicate it in the next test as well. Do you think that would be better for clarity?

spec/jobs/attach_files_to_work_job_spec.rb
bbpennel: When verifying that a method is called, you have to setup the `expect` before executing the code that would call it

spec/jobs/attach_files_to_work_job_spec.rb
lfarrell: Maybe a comment.

PR #846, opened 2022-08-25
Only pull in the method we override

lfarrell:
https://jira.lib.unc.edu/browse/HYC-1503


app/overrides/helpers/hyrax/citations_behaviors/name_behavior_override.rb
```diff
@@ -0,0 +1,19 @@
+# [hyc-override] Overriding helper in order to use creator_display field for author listings
```
bbpennel, line 1: Please link to the file that this is overriding in hyrax 2.9.6

bbpennel (changes requested):

bbpennel (approved):

PR #847, opened 2022-08-25
Move indexing override to overrides directory. Remove methods we don't override

lfarrell:
https://jira.lib.unc.edu/browse/HYC-1505

I think this if fixed in Hyrax 3 and can be removed when we upgrade, https://github.com/samvera/hyrax/blame/main/app/indexers/hyrax/deep_indexing_service.rb#L99


bbpennel (changes requested):
It looks like test coverage went down in this PR. Annoyingly, I can't tell from codeclimate which parts are not covered, but my guess is the removal of unoverridden code removed a lot of the parts that were being tested. I'm guessing we might need to add a test for the case where we fail to retrieve geonames, but running the test in docker should give a better coverage report

spec/indexers/hyrax/deep_indexing_service_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+          </gn:Feature>
+          </rdf:RDF>
+RDFXML
+
+      before do
+        stub_request(:any, "http://api.geonames.org/getJSON?geonameId=4460162&username=#{ENV['GEONAMES_USER']}").
+          with(headers: {
+                 'Accept' => '*/*',
+                 'Accept-Encoding' => 'gzip;q=1.0,deflate;q=0.6,identity;q=0.3',
+                 'User-Agent' => 'Ruby'
```
bbpennel, line 26: Since the headers aren't something we are setting ourselves, and seem to be constant, I am guessing we can leave this `with` clause out, which should help some with readability. I think we can remove the other two `with(headers` sections as well

spec/indexers/hyrax/deep_indexing_service_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+require Rails.root.join('app/overrides/indexers/hyrax/deep_indexing_service_override.rb')
+
+RSpec.describe Hyrax::DeepIndexingService, type: :indexer do
+  let(:work) { General.new(title: ['new general work']) }
+  let(:service) { described_class.new(work) }
+  subject(:solr_document) { service.generate_solr_document }
+
+  describe 'with a remote resource (based near)' do
+    context 'with a successful call to geonames' do
+      chapel_hill = <<RDFXML.strip_heredoc
```
bbpennel, line 12: Maybe name include a little more context in the variable name, like `geo_resp_chapel_hill` or something. Also wasn't sure if this should be in a `let()` block?

bbpennel (changes requested):
You might already know about this, but you can view the test coverage report on your local computer after running it in docker by going to 
`file:///path/to/hy-c/coverage/index.html`
and click through to individual files to see which lines aren't being called. It really seems like there must be a way to see that in codeclimate, but I haven't seen it yet.

app/overrides/indexers/hyrax/deep_indexing_service_override.rb
```diff
@@ -0,0 +1,38 @@
+# [hyc-override] Overriding to add method to return custom label and save it to Solr
```
bbpennel, line 1: Lets link to the file this overrides
https://github.com/samvera/hyrax/blob/v2.9.6/app/indexers/hyrax/deep_indexing_service.rb

app/overrides/indexers/hyrax/deep_indexing_service_override.rb
```diff
@@ -0,0 +1,38 @@
...
+# Note: Customizations are very specific to the Location controlled vocabulary
+Hyrax::DeepIndexingService.class_eval do
+  def geonames_mail(e)
```
bbpennel, line 4: Is it typical practice to define a method like this for sending the email rather than calling the mailer directly in the rescue block?

spec/indexers/hyrax/deep_indexing_service_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+      it 'indexes id and label' do
+        expect(solr_document.fetch('based_near_sim')).to eq ['http://sws.geonames.org/4460162/']
+        expect(solr_document.fetch('based_near_label_sim')).to eq ['Chapel Hill, North Carolina, United States']
+      end
+    end
+
+    context 'with an unsuccessful call to geonames' do
+
+      before do
+        allow(service).to receive(:geonames_mail).and_return(mail_sent)
```
bbpennel, line 48: I'd suggest doing 
`allow(GeonamesMailer).to receive(:send_mail).and_return(mail_sent)`
since this is a unit test, so it makes sense to mock an external class for this test, rather than mocking a part of the class we are testing. I'd do the same for the `expect` down below as well

spec/indexers/hyrax/deep_indexing_service_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+            headers: {
+              'Accept'=>'*/*',
+              'Accept-Encoding'=>'gzip;q=1.0,deflate;q=0.6,identity;q=0.3',
+              'User-Agent'=>'Ruby'
+            }).
+          to_return(status: 200, body: '', headers: {})
+        work.based_near_attributes = [{ id: 'http://sws.geonames.org/4460162BadId/' }]
+        stub_request(:get, 'http://sws.geonames.org/4460162BadId/').
+          with(
+            headers: {
```
bbpennel, line 61: Does the stubbing still match without specifying the expected headers? Because a human list of Accepts like this seems like it could change pretty easily when the geonames gem gets updated, and it doesn't seem very important to what we're testing.

bbpennel (changes requested):

app/overrides/indexers/hyrax/deep_indexing_service_override.rb
lfarrell: Probably not. It's called directly now.

bbpennel (approved):

PR #856, opened 2022-09-08
Hyc 1496 attribute renderer override

sharonluong:
This PR institutes an override for the attribute renderer, specifically for showing the language name instead of the language code. It also includes a test.
https://jira.lib.unc.edu/browse/HYC-1496


app/overrides/renderers/hyrax/renderers/attribute_renderer_override.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+# [hyc-override] Overriding default. Show the language term instead of the saved value.
+# Allow itemprop to be rendered
```
bbpennel, line 4: I think we removed this override

bbpennel (approved):

app/overrides/renderers/hyrax/renderers/attribute_renderer_override.rb
sharonluong: removed the comment

PR #859, opened 2022-09-12
HYC-1504: Add tests for publication behavior overrides

lfarrell:
https://jira.lib.unc.edu/browse/HYC-1504


spec/helpers/hyrax/citations_behaviors/publication_behavior_spec.rb
```diff
@@ -0,0 +1,99 @@
+# frozen_string_literal: true
+require 'rails_helper'
+require Rails.root.join('app/overrides/helpers/hyrax/citations_behaviors/publication_behavior_override.rb')
+
+RSpec.describe Hyrax::CitationsBehaviors::PublicationBehavior, type: :helper do
+  context 'with a publication date' do
+    let(:work_with_date_issued) {
```
bbpennel, line 7: In terms of test structure, I think it'd make sense for the top level to be `describe '#setup_pub_date'` and then have 4 contexts inside of it, one for each of the example articles, since they each represent a test scenario/context. Then the context info can be moved from the `it` message into the context description.

spec/helpers/hyrax/citations_behaviors/publication_behavior_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+      expect(helper.setup_pub_date(work_without_date_issued)).to eq nil
+    end
+
+    it 'does not return a formatted date for invalid dates' do
+      expect(helper.setup_pub_date(work_with_bad_date_issued)).to eq nil
+    end
+
+  end
+
+  context 'with a publication place' do
```
bbpennel, line 51: Similar suggestion to above, have a `describe '#setup_pub_place'` here, and then contexts for the two attribute sets inside of it

bbpennel (changes requested):

spec/helpers/hyrax/citations_behaviors/publication_behavior_spec.rb
lfarrell: Removed override for setup_pub_info. Local override breaks tests for https://github.com/UNC-Libraries/hy-c/blob/main/spec/helpers/hyrax/citations_behaviors/formatters/apa_formatter_spec.rb

bbpennel (approved):
I guess we care more about stray commas than hyrax does. Good work

PR #861, opened 2022-10-04
Hyc 1485 download extension override

sharonluong:
This PR moves the download behavior override for file extensions to an actual override file and adds a test for the behavior.
https://jira.lib.unc.edu/browse/HYC-1485


bbpennel (approved):
Looks good! Thanks for fixing the valid? method as well

PR #862, opened 2022-10-04
HYC-1484 - BaseActor testing/override

bbpennel:
https://jira.lib.unc.edu/browse/HYC-1484


lfarrell (changes requested):
Should honors_thesis_actor_spec.rb be named base_actor_spec.rb? It seems a bit confusing as is.

bbpennel:
I can rename it if that is helpful. I went with naming it HonorsThesisActor since BaseActor is effectively an abstract class. The equivalent test in hyrax is for GenericWorkActor:
https://github.com/samvera/hyrax/blob/v2.9.6/spec/actors/hyrax/actors/generic_work_actor_spec.rb
I chose HonorsThesisActor since the work specific permissions stuff was mostly related to honor's theses, but really it would be okay with any of the work actor's at the moment. In theory, we could probably do this without any overrides, just make a local `PersonAwarePermissionsWorkActor` that inherits from BaseActor and is used by all our work type actors. Maybe I should change this PR to that?

PR #868, opened 2022-10-13
HYC-1554 - Get work page action buttons showing again

bbpennel:
https://jira.lib.unc.edu/browse/HYC-1554

* Updates related view files to match the changes from hyrax 3
* Updates workflow presenter
* Adds cleaning to one of the tests that was failing due to the work action buttons not showing, so that the test full passes.
* Fix rubocop issues in migration files

### DIFF
--- a/app/views/hyrax/base/show.html.erb
+++ b/app/views/hyrax/base/show.html.erb
@@ -1,49 +1,62 @@
-<%# [hyc-override] %>
-<%# Remove description and display it interleaved in the metadata %>
 <% provide :page_title, @presenter.page_title %>
 
 <%= render 'shared/citations' %>
 
 <div class="row work-type">
-  <div class="col-xs-12">
+  <div class="col-sm-12">
     <%= render 'work_type', presenter: @presenter %>
   </div>
-  <div class="col-xs-12">&nbsp;</div>
   <div itemscope itemtype="http://schema.org/CreativeWork" class="col-xs-12">
+    <%= render 'work_title', presenter: @presenter %>
+    <%= render 'show_actions', presenter: @presenter %>
     <div class="panel panel-default">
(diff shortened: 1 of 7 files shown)


app/views/hyrax/base/show.html.erb
```diff
@@ -22,28 +20,54 @@
...
           <div class="col-sm-3 text-center">
+            <%# [hyc-override] Remove restriction on showing representative media info when iiif viewer present %>
             <%= render 'representative_media', presenter: @presenter, viewer: false %>
+            <%# [hyc-override] Changes end here %>
             <%= render 'citations', presenter: @presenter %>
             <%= render 'social_media' %>
           </div>
           <div class="col-sm-9">
+            <%# [hyc-override] Remove description and display it interleaved in the metadata %>
+            <%# <%= render 'work_description', presenter: @presenter %>
```
sharonluong, line 38: what is this commented out line here for?

bbpennel: This line exists in the hyrax version of the file
https://github.com/samvera/hyrax/blob/v3.4.2/app/views/hyrax/base/show.html.erb#L27
I figured it'd make it a bit easier to compare the files in the future to figure out what we changed if I left in the deleted line commented out. But I could could remove it and just leave the hyc-override tag as the reminder

sharonluong: oh! i guess it's not obvious at first glance that the commented out portion is the hyrax code. maybe a little note would suffice?

bbpennel: I updated the override message to indicate that the commenting out is related to the override, does that help?

sharonluong (approved):

PR #869, opened 2022-10-13
Setup UniversalViewer

lfarrell:
Mostly works now. We'll need to install nodejs 14 and yarn for this to work. If we use nodejs as our runtime we can get rid of miniracer as well.


lfarrell:
https://jira.lib.unc.edu/browse/HYC-1551

bbpennel (approved):

PR #871, opened 2022-10-14
HYC-1198 - Update file_actor for hyrax 3, and switch it to an override

bbpennel:
Relates to https://jira.lib.unc.edu/browse/HYC-1198


app/overrides/actors/hyrax/actors/file_actor_override.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+# https://github.com/samvera/hyrax/blob/v3.4.2/app/actors/hyrax/actors/file_actor.rb
+Hyrax::Actors::FileActor.class_eval do
+
+  alias_method :original_ingest_file, :ingest_file
+  def ingest_file(io)
```
sharonluong, line 6: what does `io` mean?

sharonluong (approved):

PR #882, opened 2022-10-18
HYC-1198 - Update Hyrax::Controller for hyrax 3

bbpennel:
Part of https://jira.lib.unc.edu/browse/HYC-1198
* I was not able to switch this to an override, since it is a module and includes a `included` block, which cannot be duplicated/overridden. Removing the unmodified code also did not work.

### DIFF
--- a/app/controllers/concerns/hyrax/controller.rb
+++ b/app/controllers/concerns/hyrax/controller.rb
@@ -1,6 +1,5 @@
 # frozen_string_literal: true
-# [hyc-override] added create_unrestricted_work_presenter as a replacement for deprecated create_work_presenter
-# [hyc-override] Add begin/rescue for locale setting
+# https://github.com/samvera/hyrax/blob/v3.4.2/app/controllers/concerns/hyrax/controller.rb
 module Hyrax::Controller
   extend ActiveSupport::Concern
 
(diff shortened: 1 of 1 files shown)


app/controllers/concerns/hyrax/controller.rb
```diff
@@ -85,10 +129,18 @@ def deny_access_for_current_user(exception, json_message)
...
-    session['user_return_to'.freeze] = request.url
+    session['user_return_to'] = request.url
     respond_to do |wants|
       wants.html { redirect_to main_app.new_user_session_path, alert: exception.message }
       wants.json { render_json_response(response_type: :unauthorized, message: json_message) }
     end
   end
+
+  # Redirect all deposit and edit requests with warning message when in read only mode
+  def check_read_only
```
sharonluong, line 140: where this this part come from? part of hyrax 3?

bbpennel: Yes, it was added in hyrax 3
https://github.com/samvera/hyrax/blob/main/app/controllers/concerns/hyrax/controller.rb#L139
For these hyrax 3 PRs, my goal is to merge in changes from hyrax rather than introduce new behaviors.

lfarrell: I added it to hyrax proper. You're welcome. :)

sharonluong: ah i see!

sharonluong (approved):

PR #892, opened 2022-10-25
HYC-1558 - Update SolrMigartionService to work with hyrax 3

bbpennel:
https://jira.lib.unc.edu/browse/HYC-1558

* Update counts in solrmigration spec to take into account new embargo and lease objects
* Adjust indexing order to place adminsets before works/filesets, without doing so tests were flapping with regards to the admin_set_tesim field
* Some rubocop fixes

### DIFF
--- a/spec/services/tasks/solr_migration_service_spec.rb
+++ b/spec/services/tasks/solr_migration_service_spec.rb
@@ -6,6 +6,8 @@ require 'fileutils'
 # https://github.com/samvera/hyrax/issues/5902
 # if this is resolved, counts will need to be readjusted
 RSpec.describe Tasks::SolrMigrationService do
+  TOTAL_SOLR_COUNT = 11
+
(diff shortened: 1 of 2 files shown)


spec/services/tasks/solr_migration_service_spec.rb
```diff
@@ -68,7 +71,7 @@
 
     it 'Lists ids and reindexes them' do
       initial_records = ActiveFedora::SolrService.get('*:*', rows: 100, sort: 'id ASC')
-      expect(initial_records['response']['numFound']).to eq 7
+      expect(initial_records['response']['numFound']).to eq 11
```
sharonluong, line 74: i dunno if this would really be useful but it seems like the expected value of 11 is repeated all over so maybe if you made a constant you'd only have to change one place later when the issue is resolved?

sharonluong (approved):

PR #905, opened 2022-11-01
Hyc 1552 add new work fields to tests

sharonluong:
added new fields from basic_metadata to the forms tests
https://jira.lib.unc.edu/browse/HYC-1552


sharonluong:
ugh meant to make this off hyrax 3

PR #914, opened 2022-11-18
HYC-1532 - Honors theses workflow, first pass

bbpennel:
https://jira.lib.unc.edu/browse/HYC-1532

* Overriding test to see if there are workflow restrictions to return false if the user has view permissions and has rights within the appropriate workflow
* Permitting visibility of fileset if user has read permissions to a suppressed object
* Merged role permissions assignment with notification, and removed extra setting of permissions on work
* Updated display_media_download_link to include our special cases rather than repeating them in each partial view
* Revoke depositor permissions on deposit in honors workflow


app/overrides/helpers/hyrax/workflows_helper_override.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+# https://github.com/samvera/hyrax/blob/v3.4.2/app/helpers/hyrax/file_set_helper.rb
+Hyrax::WorkflowsHelper.module_eval do
+  def workflow_restriction?(object, ability: current_ability)
+    return false if object.nil? # Yup, we may get nil, and there's no restriction on nil
```
lfarrell, line 5: Spacing is a little off

app/overrides/helpers/hyrax/file_set_helper_override.rb
```diff
@@ -0,0 +1,30 @@
...
+  # @param file_set [#id]
+  #
+  # @return [Boolean] whether to display the download link for the given file
+  #   set
+  # [hyc-override] reworked to include admin permissions and workflow adjustments
+  def display_media_download_link?(file_set:)
+    # Make sure download links are enabled for the app
+    return false if !Hyrax.config.display_media_download_link?
+    # Always permit admin users
+    return true if current_user&.admin?
```
lfarrell, line 16: Should this be first, so admins always have download ability?

app/services/hyrax/workflow/assign_reviewer_by_affiliation.rb
```diff
@@ -9,23 +9,39 @@ def self.call(target:, **)
...
+                         "#{department}_reviewer",
+                         'group',
+-                        'read')
       end
     end
   end
 
+  def self.notify_reviewers(work, work_id, group_name, type, access)
+    # Send notification to reviewer group
+    entity = Sipity::Entity.where(proxy_for_global_id: work.to_global_id.to_s).first
```
lfarrell, line 31: Could probably move this into the conditional to avoid loading if there are no users.

lfarrell (changes requested):

app/overrides/helpers/hyrax/file_set_helper_override.rb
bbpennel: I had it second since it made sense to me to check if download links were enabled at all before permitting them for admins.

app/overrides/helpers/hyrax/workflows_helper_override.rb
bbpennel: Thanks, should be fixed now

app/services/hyrax/workflow/assign_reviewer_by_affiliation.rb
bbpennel: Good point, I've refactored things a little to address this.

lfarrell (approved):

PR #916, opened 2022-11-28
Fix already initialized constant warning

lfarrell:
https://jira.lib.unc.edu/browse/HYC-1578


app/overrides/services/hyrax/microdata_override.rb
```diff
@@ -2,10 +2,9 @@
 # https://github.com/samvera/hyrax/blob/v3.4.2/app/services/hyrax/microdata.rb
 Hyrax::Microdata.class_eval do
   # [hyc-override] Check for local schema.org file, so our overrides can take effect.
-  local_schema_file = Rails.root.join('config', 'schema_org.yml')
```
bbpennel, line 5: We could probably have made a smaller change and done
```
local_schema_file ||= Rails.root.join('config', 'schema_org.yml')
LOCAL_FILENAME ||= File.file?(local_schema_file) ? local_schema_file : FILENAME
```
Since they only really need to be evaluated once. I don't really have a good sense of how often load_paths is called

bbpennel (changes requested):
It would be good to rebase this on hyrax-3 so the tests will pass

config/initializers/hyrax.rb
```diff
@@ -334,3 +334,8 @@
 
 # set bulkrax default work type to first curation_concern if it isn't already set
 Bulkrax.default_work_type = 'General' if Bulkrax.default_work_type.blank?
+
+# Load our local schema.org config instead of the default
+local_schema_file = Rails.root.join('config', 'schema_org.yml')
+local_filename = File.file?(local_schema_file) ? local_schema_file : FILENAME
```
bbpennel, line 340: I am guessing FILENAME is not defined here? Might work as `Hyrax::Microdata::FILENAME`, assuming its visible

config/initializers/hyrax.rb
```diff
@@ -334,3 +334,8 @@
...
+Hyrax::Microdata.load_paths=(local_filename)
```
bbpennel, line 341: Why is `local_filename` wrapped in `()` here? From the tests it seems like this method is just expecting to receive a string

bbpennel (changes requested):

bbpennel (approved):

PR #921, opened 2022-12-07
HYC-1583 - Fix analytics and solr config

bbpennel:
Part of https://jira.lib.unc.edu/browse/HYC-1583

* Fix for analytics config that was causing the dashboard to fail to load
* Switch solr queries to using POST since the review submissions page was producing solr URLs that were too long (this change is in the catalog_controller template in hyrax)
* Switch to using different fedora container for gh actions, since previous one stopped working. Now it is consistent with the dev vm

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,7 +29,7 @@ jobs:
       DOI_PREFIX: '10.17615'
       EMAIL_FROM_ADDRESS: 'hyraxapp@example.com'
       EMAIL_GEONAMES_ERRORS_ADDRESS: 'hyraxapp@example.com'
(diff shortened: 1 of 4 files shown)


sharonluong (approved):
it amuses me that we're changing a get to a post behind the scenes, though i see why we had to =D

bbpennel:
Yeah, I feel iffy about having a method called `get` perform a POST. I considered changing it so `solr_documents` (the method that calls `get` in this case) called a `post` method instead, but stuck with the smaller change for now since there isn't currently a `post` method in that class, solr doesn't really care if it gets GET or POST for queries, and there are probably other queries that go to that class that might have the same problem.

I opened a hyrax ticket related to the problem, if I make an upstream patch I might try to make it read a bit more sensibly.

PR #929, opened 2023-01-18
HYC-1602 - Convert source identifier to a single value

bbpennel:
https://jira.lib.unc.edu/browse/HYC-1602

* Override to convert source identifier to a single value, so that the source field doesn't get mangled when round tripping


bbpennel:
Fixed as part of https://github.com/UNC-Libraries/hy-c/pull/934

PR #940, opened 2023-01-27
HYC-1607/1612 - Hyrax 3 bug fixes

bbpennel:
https://jira.lib.unc.edu/browse/HYC-1607
https://jira.lib.unc.edu/browse/HYC-1612

* Skip formatting authors if the list contains only blanks, which occurs when the only creator has the name "-" (and probably other punctuation only names). Otherwise, the citation causes the record page to error out with "No implicit conversion of nil to String"
* Add back in overrides for proxy deposit requests, so that ownership transfer actions have all the necessary variables passed into the message.
* Skip adding domain suffix to system user email address when it already contains a domain (this was causing ownership transfers to fail on qa)
* Fix text color of download link when it is in a drop down
* Log a warning rather than throwing an error when depositor stats can't find a user, so that the whole statistics page doesn't blow up


spec/helpers/hyrax/citations_behaviors/formatters/chicago_formatter_spec.rb
```diff
@@ -18,6 +18,21 @@
...
+      let(:article) {
+        Article.new(title: ['new article title'],
+                    creators_attributes: { '0' => { 'name' => '-' } },
+                    date_issued: '2019-10-11',
+                    publisher: ['a publisher'],
+                    place_of_publication: ['NC'],
+                    doi: 'doi.org/some-doi')
+      }
+
+      it 'returns citation in chicago format when there is no creators' do
```
cazzerson, line 32: verb/noun aggreement

cazzerson (approved):
This is great--thank you. Just left the tiniest grammatical comment.

PR #941, opened 2023-01-30
HYC-1613 - Error showing work with no depositor in dashboard

bbpennel:
https://jira.lib.unc.edu/browse/HYC-1613

* When no depositor set for a work, user_is_depositor? will return false instead of an error so that the dashboard work page still functions


sharonluong (approved):
clever use of syntax!

PR #948, opened 2023-02-10
HYC-1628 - Fix missing work-id on file-set-in-work-download events

bbpennel:
https://jira.lib.unc.edu/browse/HYC-1628

* Include all the same data in the various download links as are in the fileset list download link


bbpennel:
Since we have other overrides in the same files, we would still need to keep these changes after the hyrax PR is merged (assuming it is) and released, but we could remove the `[hyc-override]` tags

PR #951, opened 2023-02-13
HYC-1630 - Use DownloadIR in place of file-set-in-work-download for display

bbpennel:
https://jira.lib.unc.edu/browse/HYC-1630

* Present DownloadIR stat for work downloads instead of file-set-in-work-download, since DownloadIR records all downloads (not just via the UI) and it includes stats from hyrax 2 and 3

### DIFF
--- a/app/overrides/services/hyrax/analytics/google/events_override.rb
+++ b/app/overrides/services/hyrax/analytics/google/events_override.rb
@@ -1,0 +1,6 @@
+# frozen_string_literal: true
+# https://github.com/samvera/hyrax/blob/hyrax-v3.5.0/app/services/hyrax/analytics/google/events.rb
+Hyrax::Analytics::Google::Events.module_eval do
+  # [hyc-override] use DownloadIR instead of file-set-in-work-download for download count
+  filter(:file_set_in_work_download) { |_event_action| matches(:eventAction, 'DownloadIR') }
+end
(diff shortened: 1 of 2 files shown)


app/overrides/services/hyrax/analytics/google/events_override.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+# https://github.com/samvera/hyrax/blob/hyrax-v3.5.0/app/services/hyrax/analytics/google/events.rb
+Hyrax::Analytics::Google::Events.module_eval do
+  # [hyc-override] use DownloadIR instead of file-set-in-work-download for download count
+  filter(:file_set_in_work_download) { |_event_action| matches(:eventAction, 'DownloadIR') }
```
lfarrell, line 5: is DownloadIR an event set in Google Analytics itself?

lfarrell (approved):

app/overrides/services/hyrax/analytics/google/events_override.rb
bbpennel: Its a custom event we emit to google analytics from a server side override
https://github.com/UNC-Libraries/hy-c/blob/ebbe8873e188f4d389d3e86436345bc9c694ebed/app/controllers/concerns/hyc/download_analytics_behavior.rb#L17
Looks like Jennifer added it, which turns out to have been a pretty good long term decision

PR #974, opened 2023-05-23
HYC-1367 - Proquest/sage ingests via dashboard

bbpennel:
https://jira.lib.unc.edu/browse/HYC-1367

* Adds buttons for triggering ingest of all proquest or all sage packages in the dashboard
* Adds a page for viewing the status of these ingests
* Adds sidekiq jobs for running these ingests in the background
* Refactors proquest/sage ingest services to keep track of the status of each package so that they can be presented
* Refactors ingest services to throw exceptions rather than return early
* Renames "provider" concept to "source", since this is the language we were already using in the ingest services


app/views/ingest_from_ftp/view_status.html.erb
```diff
@@ -0,0 +1,60 @@
...
+            <% @status_results.each_with_index do |(filename, entry), index| %>
+              <tr>
+                <td>
+                  <%= filename %>
+                </td>
+                <td>
+                  <%= entry['status'] %>
+                </td>
+                <td>
+                  <% if !entry['error'].nil? %>
```
bbpennel, line 30: use `present?`

spec/services/tasks/ingest_service_spec.rb
bbpennel, line 1: Delete? Seems redundant with the proquest and sage ingest service tests

app/services/tasks/ingest_status_service.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+# Service used for storing information about the status of packages being submitted for ingest
+module Tasks
+  class IngestStatusService
+    attr_reader :statuses
+
+    def initialize(status_path)
```
bbpennel, line 7: Note that the file doesn't get created immediately

sharonluong (approved):

PR #978, opened 2023-06-05
HYC-1362 - Sage Revisions

bbpennel:
https://jira.lib.unc.edu/browse/HYC-1362

* Revisions of sage articles will now update an existing work with the same Sage provided DOI
    * uses the manifest.xml file to determine whether to update the metadata+xml file, the pdf file, or both
    * if a revision is received but no existing work matches its DOI, then it will be ingested as a new work
    * Clears existing creators on updated works, but the people objects live on in fedora.
    * some light refactoring to allow for reuse between creating new works and updating.
* Throw error when unable to unzip a package rather than logging and continuing.


app/services/tasks/sage_ingest_service.rb
```diff
@@ -28,31 +32,56 @@ def process_package(package_path, _index)
...
+    def construct_ingester(jats_ingest_work, unzipped_package_dir, existing_id)
+      ingester = nil
+      package_name = File.basename(unzipped_package_dir) + '.zip'
+      doi = jats_ingest_work.identifier.first
+      if existing_id.present?
+        if SageIngestService.is_revision?(package_name)
+          ingester = Tasks::SageArticleRevisionIngester.new
+          ingester.existing_id = existing_id
+        else
+          raise "Work #{existing_id} already exists with DOI #{doi}, skipping package #{package_name}" if existing_id.present?
```
bbpennel, line 54: Can remove the if

spec/services/tasks/sage_ingest_service_spec.rb
```diff
@@ -82,11 +80,139 @@
...
+          expect(status['error'][0]['message']).to eq "Package #{package_name} indicates that it is a revision, but no existing work with DOI https://doi.org/10.1177/00031348221074228 was found. Creating a new work instead."
+          expect(status['error'][0]['trace']).to be_nil
+        end
+      end
+
+      context 'revision with existing work that does not have expected files' do
+        let(:package_name) { 'ASU_2022_88_10_10.1177_00031348221074228.r2022-12-22.zip' }
+        let(:work) { FactoryBot.create(:article, identifier: ['https://doi.org/10.1177/00031348221074228']) }
+
+        it 'creates the work as if it were new' do
```
bbpennel, line 105: update this message

spec/services/tasks/sage_ingest_service_spec.rb
```diff
@@ -82,11 +80,139 @@
...
+            pdf_date_modified = pdf_fs.date_modified
+
+            expect { service.process_package('spec/fixtures/sage/revisions/metadata_changed/ASU_2022_88_10_10.1177_00031348221074228.r2022-12-19.zip', 0) }
+                .to change { Article.count }.by(0)
+                .and change { FileSet.count }.by(0)
+            work = ActiveFedora::Base.find(@work_id)
+            expect(work.title.first).to eq 'America\'s Original Immunization Controversy'
+            creators = work.creators.to_a
+            expect(creators.length).to eq 1
+            expect(creators[0].name.first).to eq 'Nakayama, Don K.'
```
bbpennel, line 151: Use constants for title and creator name to make it clearer what the original and the updated values are

sharonluong (approved):

PR #979, opened 2023-06-08
HYC-1687 - DOI script fixes

bbpennel:
https://jira.lib.unc.edu/browse/HYC-1687

* Converts languages to iso639-1, which is the required format for languages in datacite
    * Language vocab now has iso639-1 values when one exists.
* Removes contributors from datacite request as they are no longer supported by its API
* Excludes items without publication dates from receiving a DOI (it is a required field)
* Items with invalid publication dates will fall back to the date the work was created (dates must be YYYY)
* Dates that are in decade/century format will converted back an actual date (18xx becomes 1800)
* resourceTypeGeneral now falls back to 'Text' if we are unable to determine a value (it is a required field). resourceType falls back to the resourceTypeGeneral if none is available (also a required field)

### DIFF
--- a/app/services/languages_service.rb
+++ b/app/services/languages_service.rb
@@ -17,6 +17,13 @@ module LanguagesService
     nil
   end
 
+  def self.iso639_1(id)
+    authority.find(id).fetch('iso639_1')
+  rescue KeyError
+    Rails.logger.warn "DepartmentsService: cannot find '#{id}'"
+    nil
+  end
+
   def self.include_current_value(value, _index, render_options, html_options)
     unless value.blank?
       html_options[:class] << ' force-select'
--- a/app/services/tasks/doi_create_service.rb
+++ b/app/services/tasks/doi_create_service.rb
@@ -123,12 +123,7 @@ module Tasks
                                                publisher.first
                                              end
 
-      publication_year = parse_field(work, 'date_issued')
-      data[:data][:attributes][:publicationYear] = if publication_year.blank?
-                                                     Date.today.year
-                                                   else
-                                                     Array.wrap(publication_year).first.to_s.match(/\d{4}/)[0]
-                                                   end
+      data[:data][:attributes][:publicationYear] = extract_publication_year(work)
 
       ############################
       #
@@ -161,6 +156,16 @@ module Tasks
       data[:data][:attributes][:subjects] = subjects unless subjects.blank?
 
       data.to_json
+    end
+
+    def extract_publication_year(work)
+      date_issued = parse_field(work, 'date_issued')
+      year_match = Array.wrap(date_issued).first.to_s.match(/\d{4}/)
+      if year_match.nil?
+        work.create_date.year.to_s
+      else
+        year_match[0]
+      end
     end
 
     def create_doi(record)
--- a/app/services/tasks/doi_create_service.rb
+++ b/app/services/tasks/doi_create_service.rb
@@ -135,9 +135,6 @@ module Tasks
       # Optional fields
       #
       ############################
-      contributors = parse_people(work, 'contributors')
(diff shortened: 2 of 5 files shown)


app/services/languages_service.rb
```diff
@@ -17,6 +17,13 @@ def self.label(id)
     nil
   end
 
+  def self.iso639_1(id)
+    authority.find(id).fetch('iso639_1')
+  rescue KeyError
```
sharonluong, line 22: so if it can't find the iso639_1 version then it will return nil?

app/services/tasks/doi_create_service.rb
```diff
@@ -123,29 +123,24 @@ def format_data(work)
...
 
       funding = parse_funding(work, 'funder')
       data[:data][:attributes][:fundingReferences] = funding unless funding.blank?
 
-      language = parse_field(work, 'language_label').first
-      data[:data][:attributes][:language] = language unless language.blank?
+      language = parse_field(work, 'language').first
+      if language.present?
+        lang_code = LanguagesService.iso639_1(language)
+        data[:data][:attributes][:language] = lang_code unless lang_code.blank?
```
sharonluong, line 142: what does `data[:data][:attributes][:language]` get set to if lang_code is blank?

app/services/tasks/doi_create_service.rb
```diff
@@ -163,6 +158,18 @@ def format_data(work)
       data.to_json
     end
 
+    def extract_publication_year(work)
```
sharonluong, line 161: i feel like using a verb here is a java convention?

app/services/tasks/doi_create_service.rb
```diff
@@ -242,11 +250,19 @@ def parse_resource_type(dcmi_type, record_type)
         resource_type = record_type&.first
         datacite_type = RESOURCE_TYPE_TO_DATACITE[resource_type]
       end
-      puts "#{get_time} WARNING: Unable to determine resourceTypeGeneral for record" if datacite_type.nil?
+      if datacite_type.nil?
+        puts "#{get_time} WARNING: Unable to determine resourceTypeGeneral for record" if datacite_type.nil?
```
sharonluong, line 254: i think the `if datacite_type.nil?` check here isn't needed since it's checked in line 253 already

app/services/tasks/doi_create_service.rb
```diff
@@ -242,11 +250,19 @@ def parse_resource_type(dcmi_type, record_type)
...
+        datacite_type = 'Text'
+      end
+
       # Storing the datacite type. If it is nil or invalid, datacite will reject the creation
       result[:resourceTypeGeneral] = datacite_type
 
-      result[:resourceType] = record_type.first unless record_type.blank?
+      if record_type.present?
+        result[:resourceType] = record_type.first unless record_type.blank?
```
sharonluong, line 262: don't need the `unless record_type.blank?` here

app/services/tasks/doi_create_service.rb
```diff
@@ -242,11 +250,19 @@ def parse_resource_type(dcmi_type, record_type)
...
+      else
+        result[:resourceType] = datacite_type
+      end
```
sharonluong, line 265: hmm or you could use a ternary like

result[:resourceType] = record_type.present? ? record_type.first : datacite_type

sharonluong (changes requested):

app/services/tasks/doi_create_service.rb
bbpennel: nothing, no language attribute would be supplied. Its a recommended but not required field.

app/services/tasks/doi_create_service.rb
bbpennel: good catch

sharonluong (approved):

PR #982, opened 2023-06-15
HYC-1374 - CLEANUP_FTP_PACKAGES is supplied as a string

bbpennel:
Continuation of https://jira.lib.unc.edu/browse/HYC-1374

* Handle when CLEANUP_FTP_PACKAGES is supplied as a string, and break into a helper method

### DIFF
--- a/app/services/tasks/ingest_service.rb
+++ b/app/services/tasks/ingest_service.rb
@@ -98,6 +98,10 @@ module Tasks
       end
     end
 
+    def cleanup_enabled?
+      @cleanup_enabled ||= ENV['CLEANUP_FTP_PACKAGES'].to_s.downcase == 'true'
+    end
+
     def extract_files(package_path)
       dirname = unzip_dir(package_path)
       logger.info("Extracting files from #{package_path} to #{dirname}")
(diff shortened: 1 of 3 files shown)


app/services/tasks/ingest_service.rb
```diff
@@ -98,6 +98,10 @@ def ingest_progress_log
       end
     end
 
+    def cleanup_enabled?
+      @cleanup_enabled ||= ENV['CLEANUP_FTP_PACKAGES'].to_s.downcase == 'true'
```
sharonluong, line 102: this makes me a bit worried. can you use something like `ENV['CLEANUP_FTP_PACKAGES'].to_b` or maybe `ActiveModel::Type::Boolean.new.cast(ENV['CLEANUP_FTP_PACKAGES'])` (in rails 5) to cast the value as a boolean instead?

bbpennel: Thanks, I looked at `ActiveModel::Type::Boolean.new.cast` but decided against it mostly because it defaults to TRUE if the value doesn't match any versions of `false` it expects (which are case sensitive), which isn't the behavior I'd want for an invalid value. I can't find `.to_b` in regular ruby or in rails, so i'm not sure about that one

sharonluong: omg i was looking at this but i guess it's a ruby library?? https://www.rubydoc.info/github/rubyworks/facets/String:to_b

sharonluong: ok, i guess after all this, the code looks fine :)

sharonluong (approved):

PR #984, opened 2023-06-29
Hyc 1694 delete filesets bug

sharonluong:
This PR makes it so that only admins can see the option to delete a fileset, either in a dropdown or on the filesets show page. It also adds an admin check to the controller so that backend attempts to delete must be by an admin as well.

https://jira.lib.unc.edu/browse/HYC-1694


app/views/hyrax/file_sets/_show_actions.html.erb
```diff
@@ -0,0 +1,21 @@
+<%# [hyc-override] Overriding to make only admins able to delete a fileset %>
+<div class="form-actions">
+  <% if Hyrax.config.analytics? %>
```
bbpennel, line 3: I think the base version of this file may be based off of hyrax main rather than hyrax 3.6.0 (or 3.x in general) since I'm noticing a number of differences in this copy of the file versus 3.x, and I think that is why some of the buttons on the fileset page are no longer styled. Could you reimport this file from 3.6.0 and reapply your changes?
https://github.com/samvera/hyrax/blob/hyrax-v3.6.0/app/views/hyrax/file_sets/_show_actions.html.erb
That will also get the change in 3.6.0 that disables this analytics link, which we want to make sure is included.

app/views/hyrax/file_sets/_actions.html.erb
```diff
@@ -0,0 +1,54 @@
+<%# [hyc-override] Overriding to see if user can delete the work, to be able to delete the fileset %>
```
bbpennel, line 1: I think this override is based off of hyrax main rather than 3.x, could you reimport the base version of the file from 3.6.0 and reapply your changes?
https://github.com/samvera/hyrax/blob/hyrax-v3.6.0/app/views/hyrax/file_sets/_actions.html.erb

bbpennel (changes requested):
It looks good over all, but I left two comments related to making sure we are overriding files from the version of hyrax we are using.

app/views/hyrax/file_sets/_show_actions.html.erb
sharonluong: done!

app/views/hyrax/file_sets/_actions.html.erb
sharonluong: done!

bbpennel (approved):

PR #985, opened 2023-07-05
HYC-1699 - Analytics cache script

bbpennel:
https://jira.lib.unc.edu/browse/HYC-1699

* Adds a script that pulls data from GA3 (UA) for every work and fileset in the repository, populating the local database cache
* Performs this work in parallel
* Adds analytics.yml config back in for GA3, so that the application can be used with GA3 or 4


sharonluong (approved):
Nice work!

PR #986, opened 2023-07-11
Hyc 1686 manually delete package HYC 1685 select ingest packages

sharonluong:
https://jira.lib.unc.edu/browse/HYC-1686
https://jira.lib.unc.edu/browse/HYC-1685

This PR introduces check boxes to the ingest FTP packages page. Now you may select packages you either wish to delete or ingest. There is also a select all checkbox to make it easier to delete all or ingest all packages.


app/controllers/ingest_from_ftp_controller.rb
```diff
@@ -32,12 +41,40 @@ def view_status
...
+    source
+    selected_filenames = params[:selected_filenames]
+    if selected_filenames.blank?
+      flash[:alert] = 'No packages were chosen'
+      redirect_to ingest_from_ftp_path(source: @source)
+      return
+    end
+    package_paths = list_package_files
+    package_paths.each do |package_path|
+      if selected_filenames.any? { |filename| package_path.include?(filename) }
```
bbpennel, line 55: Another approach could be to loop through `selected_filenames` and delete `File.join(storage_base_path, File.basename(selected_filename))` if it exists (with the basename part to make sure the user doesn't include as "../" in the filename or try to change directories).

If we want to keep the current approach, would it make sense to call `list_selected_package_paths` and delete everything in that instead? It looks like it would only include filenames that exist in the sage/proquest directory.

app/controllers/ingest_from_ftp_controller.rb
```diff
@@ -32,12 +41,40 @@ def view_status
...
   private
 
   def list_package_files
     Dir[File.join(storage_base_path, '*.zip')]
   end
 
+  def list_selected_package_paths(selected_filenames)
+    selected_package_paths = []
+    list_package_files.each do |package_path|
+      if selected_filenames.any? { |filename| package_path.include?(filename) }
```
bbpennel, line 71:
```suggestion
      if selected_filenames.any? { |filename| File.basename(package_path) == filename }
```
since we're only concerned with the filename part and it'd eliminate the very small chance of filename matching only part of the path (either in the parent directories or a partial filename match).

app/views/ingest_from_ftp/list_packages.html.erb
```diff
@@ -52,6 +58,19 @@
...
   </div>
 </div>
+<script>
+  $(document).ready(function(){
+    $('#select_all').on("click", function() {
+      var checkboxes = $(".checkbox");
+      if(checkboxes.prop("checked")){
+        checkboxes.prop("checked",false);
+      } else {
+        checkboxes.prop("checked",true);
```
bbpennel, line 72:
```suggestion
      if (checkboxes.prop("checked")){
        checkboxes.prop("checked", false);
      } else {
        checkboxes.prop("checked", true);
```

spec/controllers/ingest_from_ftp_controller_spec.rb
```diff
@@ -83,21 +83,38 @@
...
         expect(response).to redirect_to(ingest_from_ftp_status_path(source: 'sage'))
       end
+
+      it 'ingests a selection of packages and goes to the status page' do
+        post :ingest_packages, params: { source: 'proquest', selected_filenames: ['etdadmin_upload_3806.zip', 'etdadmin_upload_942402.zip']},
+          session: valid_session
+          expect(IngestFromProquestJob).to have_received(:perform_later) do |arg1, arg2|
+            expect(arg1).to eq(user.uid)
+            expect(arg2).to match_array([proquest_package1, proquest_package2])
+          end
```
bbpennel, line 108:
```suggestion
        expect(IngestFromProquestJob).to have_received(:perform_later) do |arg1, arg2|
          expect(arg1).to eq(user.uid)
          expect(arg2).to match_array([proquest_package1, proquest_package2])
        end
```

spec/controllers/ingest_from_ftp_controller_spec.rb
```diff
@@ -150,4 +167,63 @@
...
   end
+
+  describe 'POST #delete_packages' do
+    context 'as an admin' do
+      before do
+        allow(controller).to receive(:authorize!).with(:read, :admin_dashboard).and_return(true)
+        allow(controller).to receive(:current_user).and_return(user)
+      end
+
+      it 'deletes a proquest ingest job and returns to the same page' do
```
bbpennel, line 178:
```suggestion
      it 'deletes a proquest ingest package and returns to the same page' do
```

spec/controllers/ingest_from_ftp_controller_spec.rb
```diff
@@ -150,4 +167,63 @@
...
+
+    context 'as a non-admin' do
+      it 'returns an unauthorized response' do
+        post :delete_packages, params: {}, session: valid_session
+        expect(response).to redirect_to new_user_session_path
+      end
+    end
+
+    def deleted_files_exist?(filename_array, deleted_filenames)
+      existing_filenames = filename_array.map { |path| path.split('/').last }
```
bbpennel, line 223:
```suggestion
      existing_filenames = filename_array.map { |path| File.basename(path) }
```

spec/controllers/ingest_from_ftp_controller_spec.rb
```diff
@@ -150,4 +167,63 @@
...
+        expect(deleted_files_exist?(package_filepaths, selected_filenames)).to eq false
+      end
+
+      it 'deletes an array of packages and returns to the same page' do
+        post :delete_packages, params: { source: 'proquest', selected_filenames: ['etdadmin_upload_3806.zip', 'etdadmin_upload_942402.zip']},
+          session: valid_session
+        expect(response).to redirect_to(ingest_from_ftp_path(source: 'proquest'))
+
+        package_filepaths = Dir[File.join(proquest_dir.to_s, '*.zip')]
+        expect(package_filepaths.empty?).to eq true
```
bbpennel, line 204:
```suggestion
        expect(package_filepaths).to be_empty
```
Just a handy built in check, same outcome of course

bbpennel (changes requested):
Looks good based on reading, just some suggestions for tidying up

app/controllers/ingest_from_ftp_controller.rb
sharonluong: I like the second approach. When I did the delete feature I didn't have that method yet, I made it as part of the ingest selected package work.

bbpennel (approved):

PR #988, opened 2023-07-14
HYC-1712 - Restore "Ingest from FTP" and "Assign Admin Set Worktype" menu entries

bbpennel:
https://jira.lib.unc.edu/browse/HYC-1712

* Remove repository_contents partial override since bulkrax conflicts with it and takes precedence. 
* Moved Ingest from FTP and Assign Admin Set Worktype down to the Tasks menu


sharonluong (approved):
thanks for fixing!

PR #989, opened 2023-07-21
HYC-1718 - Honors thesis approval permissions bug

bbpennel:
https://jira.lib.unc.edu/browse/HYC-1718

* Adds retry for sipity additional actions when a ModelMismatch occurs, since they appears to be transient while permissions of children are being modified in Fedora
* Add back in fedora permissions update.
* Remove weird stray -


spec/services/hyrax/workflow/workflow_action_service_spec.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
```
lfarrell, line 1: Does it need a test for when there's no comment?

lfarrell (approved):

PR #996, opened 2023-09-07
HYC-1739 - Blacklight class overrides

bbpennel:
Addresses most of https://unclibrary.atlassian.net/browse/HYC-1739

* Remove override of facets_helper_behavior since we don't display individual date values in the facet, so there's no reason to format them
* Converts our dublin_core document class into an override (as a result, no updates needed to be pulled in from upstream)
    * Adds test coverage of our behavior modifications
* Deleting solr_document_wrapper.rb override, the override was no longer needed since upstream they addressed the security issue where you could retrieve private records via oai get https://github.com/projectblacklight/blacklight_oai_provider/pull/27

Does not update the following blacklight related files, I'm deferring that work to later once more of the UI is functional since they may all need to be reimplemented (details are listed on the ticket):
* app/helpers/range_limit_helper.rb
* app/views/catalog/_constraints.html.erb
* app/views/catalog/_facet_pagination.html.erb


spec/models/concerns/blacklight/document/dublin_core_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+                    translators_attributes: { '0' => { name: 'User4, A',
+                                                      other_affiliation: 'Oregon Health & Science University',
+                                                      index: 2 },
+                                             '1' => { name: 'User1, B',
+                                                      affiliation: 'Eshelman School of Pharmacy, Division of Pharmaceutical Outcomes and Policy',
+                                                      index: 1 } }
+        )
+      }
+
+      it 'returns xml document with contributors in index order, and no affiliation' do
```
lfarrell, line 72: Is this actually checking that the affiliation(s) aren't present?

spec/models/concerns/blacklight/document/dublin_core_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+      end
+    end
+
+    context 'with thumbnail' do
+      let(:work) { Article.new(id: '123456', title: ['654321']) }
+      before do
+        work_solr['thumbnail_path_ss'] = '/downloads/123456?file=thumbnail'
+      end
+
+      it 'returns xml document with contributors in index order, and no affiliation' do
```
lfarrell, line 101: This test has the same `it` block as the test on line 72. Should it be something else?

spec/models/concerns/blacklight/document/dublin_core_spec.rb
bbpennel: Affiliations currently get added as dc:contributors for creators (see the test above this one), while for fields that map to contributors (like translator) the affiliation is not added to the DC at all. We're going to be completely removing affiliation from the DC mapping soon, but i decided to wait to do that in another PR rather than change behavior here. I could clarify the message here if that'd be useful

spec/models/concerns/blacklight/document/dublin_core_spec.rb
bbpennel: oops, I'll fix that

PR #997, opened 2023-09-07
HYC-1731 - Remove access and use fields from solr document

bbpennel:
Part of https://unclibrary.atlassian.net/browse/HYC-1731

* Removes from solr document that we already removed from all the models.
* Update webdrivers and list of expected download paths for chromium so unrelated feature tests will pass again.


krwong (approved):
Looks good!

PR #998, opened 2023-09-15
HYC-1736 - Update view overrides based on hyrax 4

bbpennel:
https://unclibrary.atlassian.net/browse/HYC-1736

* Updates all .html.erb that are overridden from the hyrax gem where there were updates between hyrax 3 and 4
* Updates to javascript includes to fix blacklight errors, and more closely match source file to make comparison easier
* Some fixes for style of the homepage, but it still has issues
* Tried to clarify where overrides are and identified some that weren't previously
* Attempted to update citation modal to work with new styles/bootstrap
* Replaced dashboard menu overrides by adding menu partials, and using CSS to hide menu entries


app/views/hyrax/base/_currently_shared.html.erb
```diff
@@ -11,8 +11,8 @@
   <tr id="file_permissions">
     <td>
       <%# [hyc-override] Overriding to make depositor a facet search and not link to user profile %>
-      <%= label_tag :owner_access, class: "control-label" do %>
-        <%= t('.depositor') %> (<span id="file_owner" data-depositor="<%= permission_service.depositor %>"><%= link_to permission_service.depositor, main_app.search_catalog_path(f: { depositor_ssim: [permission_service.depositor]}) %></span>)
+      <%= label_tag :owner_access, class: "col-form-label" do %>
+        Depositor (<span id="file_owner" data-depositor="<%= permission_service.depositor %>"><%= link_to permission_service.depositor, main_app.search_catalog_path(f: { depositor_ssim: [permission_service.depositor]}) %></span>)
```
lfarrell, line 15: Did the translation for this go away?

lfarrell (approved):

app/views/hyrax/base/_currently_shared.html.erb
bbpennel: It wasn't in the original hyrax file (including if i go back through its history):
https://github.com/samvera/hyrax/blob/hyrax-v4.0.0/app/views/hyrax/base/_currently_shared.html.erb#L13C1-L13C1
I don't see a depositor field under currently shared (i think this is where it would go?):
https://github.com/samvera/hyrax/blob/hyrax-v4.0.0/config/locales/hyrax.en.yml#L497
and i don't see one in our local hyrax.en.yml, so I'm not totally sure where the value was coming from. I don't mind adding it back in, but I'm just not sure what it was doing

PR #1002, opened 2023-09-20
HYC-1740 - Hyrax 4 updates to overrides from other hydra gems (mostly)

bbpennel:
https://unclibrary.atlassian.net/browse/HYC-1740

* Converts a few files to override format, and adds tests for the overrides
* Updates to newer version of mini_magick so we could reduce our override of info.rb some
* Deletes some view overrides that were just there to get i18n text to display correctly, since it seems to be behaving alright without them now
* Pull in a few changes from upstream
* Pulls in updates from main to prevent chrome driver failures


spec/lib/hydra/role_management/user_roles_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+
+    before do
+      user.roles << librarian
+    end
+
+    it 'returns true for existing group name' do
+      expect(subject.admin_unit_manager?('librarian')).to be_truthy
+    end
+
+    it 'returns true for non-existent group' do
```
lfarrell, line 20: Same name as test above

lfarrell (changes requested):

PR #1005, opened 2023-09-25
HYC-1752 - Hyrax 4 CSS updates

lfarrell:
Doesn't include issue 11 from the ticket, as I can't currently create any works to view what the file option dropdown looks like.

### DIFF
--- a/app/assets/stylesheets/unc_custom.css.scss
+++ b/app/assets/stylesheets/unc_custom.css.scss
@@ -491,23 +491,10 @@ Resets it to the default Bootstrap card margins
   width: 100%;
 }
 
-#documents {
-  display: inline-flex;
-  flex-wrap: wrap;
-  justify-content: center;
-  margin: 0;
-
-  .document {
-    border-bottom: inherit;
-  }
-  .masonry.document {
-    background-color: white;
-    border: 1px solid $light-gray;
-    margin: 0 10px 10px 0;
-  }
-  fieldset {
-    width: 100%;
-  }
+.search-result-wrapper {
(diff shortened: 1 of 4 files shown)


bbpennel:
It's looking a lot better, there are two parts that seem like they could use fixes though:

1. On the advanced search, the "AND have these attributes" fields aren't formatted correctly compared to prod
2. On the homepage, the blue background in the "Deposit Your Work" still cuts off before the sides of the window

There were two other things not reported on the ticket that I'm noticing now. Some of them are probably out of scope for this ticket, but they are:
1. Also for the "Deposit Your Work" section, i'm noticing there's supposed to be a drop shadow below the big picture on the blue area, but on this branch its only showing the shadow on the exposed white background at the edges.
2. On the admin dashboard, all of the pie graphs (visibility, work types, resource types) are being duplicated on top of each other.
3. On the search results page none of the buttons at the top of the results are showing (like gallery view, sorting, etc), but that's probably out of scope for this ticket

lfarrell:
In the advanced search the "AND" attribute options don't show up at all. It should be noted that the file we override _advanced_search_form_html.erb where these are supposed to show up doesn't exist in BL advanced search 8. It doesn't seem like BL advanced search 8 does much of anything. In fact, this gem may go away entirely. It's still being debated in the wider community.

I don't have any works so I can't see the the graphs or the search results buttons. There was this regarding the search result buttons: https://github.com/samvera/hyrax/commit/7a2d11b3aed7d626cd8171dc1e9d4812be5c37d4. I'm not sure why they were removed. I've added them back in.

bbpennel (approved):

PR #1035, opened 2023-10-24
Hyc-1788/1794 - Sidebar select color, and user links dropdown fixes (hyrax-4)

bbpennel:
https://unclibrary.atlassian.net/browse/HYC-1794 (addressed by https://github.com/samvera/hyrax/pull/6382 in our local fork)
https://unclibrary.atlassian.net/browse/HYC-1788 (addressed upstream by https://github.com/samvera/hyrax/pull/6381 but we have an override of the file and some conflicting CSS, so the change is duplicated here)


lfarrell (approved):
This will probably need a hy-c gems update as well, since the aws-partitions gem changed.

PR #1049, opened 2023-11-13
HYC-1812 - Collection page masonry view errors

bbpennel:
https://unclibrary.atlassian.net/browse/HYC-1812

* Comment out CatalogController lines which were setting up bookmark behaviors, since we have disabled the routes for it. Removing these lines resolves `undefined method `bookmark_path'` errors directly, and seem to indirectly resolve the original `guest_uid_authentication_key` error. See the ticket for more details.

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -81,15 +81,16 @@ class CatalogController < ApplicationController
     config.index.thumbnail_field = 'thumbnail_path_ss'
     config.index.constraints_component = Hyc::ConstraintsComponent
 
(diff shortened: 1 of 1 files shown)


sharonluong (commented):
is there a reason why we can't just delete these lines?

bbpennel:
They get generated as part of the base version of CatalogController, so I guess I left them in to make comparing against the default easier. I can remove them if you think they should go though.

sharonluong:
@bbpennel ah ok that makes sense then

sharonluong (approved):

PR #1071, opened 2024-01-12
HYC-1351 Switch CC licenses from 3.0 to 4.0

davidcam-src:
https://unclibrary.atlassian.net/browse/HYC-1351

- included 4.0 CC licenses to license definitions
- marked 3.0 CC license definitions as "archived"
- filter added to CdrLicenseService so that only 4.0 licenses are available for selection by non-admins
- archived and non archived licenses are still available for selection by admins


bbpennel (changes requested):
One minor fix, but otherwise it looks good. 

I noticed that the vocab in hyrax 4 and 5 (maybe that's where you copied it from?) has https URLs, so it does seem like there's something different happening with our app. I double checked the base version of hyrax 5, and it seemed to do alright with the https vocab entries listed in their license.yml. I dug around a little but nothing is standing out to me. We're on the same version of QuestioningAuthority, nothing obvious for converting https to http. We've overridden the license related form elements to use our own license service rather than using the config option (which might have been added later), among a few other changes. We don't actually show the license URL in the UI, and the http URLs resolve to the https just fine anyways, so the impact of this is fairly small, so it's probably not worth digging into further, but it is mysterious.

bbpennel (approved):

PR #1076, opened 2024-02-15
HYC-1861 - Reduce logging of common errors

bbpennel:
https://unclibrary.atlassian.net/browse/HYC-1861

* Reduces logging of a number of errors from FATAL to WARN without a full stack trace
* Directs additional errors to appropriate error pages
* Ensures 404 errors use the correct layout, otherwise some errors (such as Ldp::Gone on an edit link) will fail since they try to use the dashboard layout, which throws an error if the user isn't logged in.
* Adds additional chromedriver url to list of real URLs allowed to be accessed during tests, since it keeps moving around and causing tests to fail
* Gets `edit_sage_ingested_works_spec.rb` to pass locally again, which was failing due to javascript not executing. This came to my attention because the test failed in ghactions due to the chromedriver issue. Refactors it to only ingest one work at a time instead of every single Sage work for each test, since otherwise the test takes 5-6 minutes to run locally.

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -9,7 +9,8 @@ require 'rspec/rails'
 require 'webmock/rspec'
 
 WebMock.disable_net_connect!(allow_localhost: true, allow: ['fedora:8080', 'solr8:8983', 'fcrepo:8080', 'solr:8983', 'opaquenamespace.org',
-                                                            'googlechromelabs.github.io', 'chromedriver.storage.googleapis.com', 'https://edgedl.me.gvt1.com'])
+                                                            'googlechromelabs.github.io', 'chromedriver.storage.googleapis.com', 'https://edgedl.me.gvt1.com',
+                                                            'storage.googleapis.com'])
 Capybara.default_normalize_ws = true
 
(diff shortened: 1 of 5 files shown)


spec/rails_helper.rb
davidcam-src, line 1: I'm not too familiar with this, would you mind explaining the change?

davidcam-src (approved):

spec/rails_helper.rb
bbpennel: https://stackoverflow.com/questions/24145329/how-is-spec-rails-helper-rb-different-from-spec-spec-helper-rb-do-i-need-it
is probably a better explanation than I can give for this file. 

In terms of the change to this file, WebMock is a utility that lets us mock http URLs in tests to return whatever response we need. In order to do its job, it catches all HTTP requests, and throws an exception if an external URL is requested unless it is either explicitly mocked or is in this `allow` list. Another testing dependency triggers download of chromedrivers in order to do testing/rendering of web pages. Google keeps moving the url where those drivers are downloaded from, so we have to keep updating the `allow` list so that webmock will allow the driver to be downloaded.

PR #1078, opened 2024-02-19
Add alt text to thumbnails

lfarrell:
https://unclibrary.atlassian.net/browse/HYC-1856

### DIFF
--- a/app/views/hyrax/admin/admin_sets/show.html.erb
+++ b/app/views/hyrax/admin/admin_sets/show.html.erb
@@ -1,0 +1,38 @@
+<%# https://github.com/samvera/hyrax/blob/hyrax-v4.0.0/app/views/hyrax/admin/admin_sets/show.html.erb %>
+<%# [hyc-override] Add alt text for thumbnails  @todo can be removed when fixed upstream %>
+<% provide :page_header do %>
+  <h1><span class="fa fa-sitemap" aria-hidden="true"></span> <%= t('.header') %></h1>
+<% end %>
+
+<div class="row">
+  <div class="col-md-12">
+    <div class="card">
+      <%= render 'show_actions', presenter: @presenter %>
+
+      <div class="card-body admin-set row">
+        <div class="col-md-2">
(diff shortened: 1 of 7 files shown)


app/views/hyrax/admin/admin_sets/show.html.erb
```diff
@@ -0,0 +1,38 @@
...
+<div class="row">
+  <div class="col-md-12">
+    <div class="card">
+      <%= render 'show_actions', presenter: @presenter %>
+
+      <div class="card-body admin-set row">
+        <div class="col-md-2">
+          <% if has_thumbnail? @presenter.solr_document %>
+            <div class="document-thumbnail">
+              <%= render_thumbnail_tag(@presenter.solr_document, { alt: '' }, { suppress_link: true }) %>
```
bbpennel, line 16: will having a blank alt text cause the check to pass? Or does it fall back to some kind of default, or should we put something like "thumbnail" in here?

app/views/hyrax/base/_file_manager_thumbnail.html.erb
```diff
@@ -0,0 +1,3 @@
+<%# https://github.com/samvera/hyrax/blob/hyrax-v4.0.0/app/views/hyrax/base/_file_manager_thumbnail.html.erb %>
+<%# [hyc-override] Add alt text for thumbnails  @todo can be removed when fixed upstream %>
+<%= document_presenter(node)&.thumbnail&.thumbnail_tag({ alt: '', class: 'thumbnail-inner mw-100' }, {}) %>
```
bbpennel, line 3: Same question about blank value

app/views/hyrax/admin/admin_sets/show.html.erb
lfarrell: I'm not sure what to put in these. According to the to the W3C an empty alt tag is valid, not having an alt tag is not. I'll just go ahead and put thumbnail in.

bbpennel (approved):

PR #1081, opened 2024-02-27
HYC-1863 - Prevent invalid filter field values

bbpennel:
https://unclibrary.atlassian.net/browse/HYC-1863

* Replace the blacklight filter field 'f' with a Parameters object if it has an invalid value in it, like a string. 
* Update blacklight version

### DIFF
--- a/spec/controllers/application_controller_spec.rb
+++ b/spec/controllers/application_controller_spec.rb
@@ -14,4 +14,22 @@ RSpec.describe ApplicationController, type: :controller do
       expect(response).to have_http_status(:request_timeout)
     end
   end
+
+  describe '#replace_invalid_f_parameter' do
+    controller(ApplicationController) do
+      def index
+        @params = params
+      end
+    end
+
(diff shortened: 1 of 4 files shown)


spec/controllers/application_controller_spec.rb
davidcam-src, line 1: Just wanted to check my understanding for the first test, since 'edit' is provided as a filter parameter and it's not an ActionController::Parameter object, we return an empty parameter object right?

Also small nitpick, I think it should be 'object' on line 30

bbpennel: right, if the value of the f field is anything other than a `ActionController::Parameters` object or nil, then it'll be replaced with an empty ActionController::Parameters object. There are many parts of the application that depend on f having a valid value (at least in blacklight, hyrax, and hyc), so I went with correcting the value rather than making the various parts of the code able to handle invalid values.

davidcam-src: okay, thanks

davidcam-src (approved):

PR #1092, opened 2024-04-04
HYC-1894 - Handle null character in download URLs

bbpennel:
https://unclibrary.atlassian.net/browse/HYC-1894

* Handle null character in file parameter and in other parts of the requested URL, which addressed "ArgumentError (path name contains null byte)" errors


davidcam-src (approved):
lgtm!

PR #1094, opened 2024-04-12
HYC-1895 - Prevent error when logged in with invalid locale

bbpennel:
https://unclibrary.atlassian.net/browse/HYC-1895

* Change the locale param too when an invalid value is provided, so that other parts of the code base that directly access it rather than `I18n.locale` won't receive an invalid locale, such as render_notifications


davidcam-src (approved):
looks good!

PR #1098, opened 2024-05-14
HYC-1896 - Decode equal signs in query string before parsing params

bbpennel:
https://unclibrary.atlassian.net/browse/HYC-1896

Add middleware which decodes equal signs in query strings prior to parameter parsing, to prevent failures when the equal sign separator between the param name and value is encoded

For example, prior to this PR this request fails:
`/catalog?f%5Bkeyword_sim%5D%5B%5D%3DPharmacogenetics&locale=en`

This PR results the query string becoming:
`/catalog?f%5Bkeyword_sim%5D%5B%5D=Pharmacogenetics&locale=en`
So that param parsing will correctly be able to parse out the keyword_sim = Pharmacogenetics search

### DIFF
--- a/app/middleware/decode_query_string.rb
+++ b/app/middleware/decode_query_string.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+class DecodeQueryString
+  def initialize(app)
+    @app = app
+  end
+
+  def call(env)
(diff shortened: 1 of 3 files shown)


davidcam-src (approved):
looks good!

PR #1100, opened 2024-05-25
HYC 1678 - Dimensions ingest email report generation

davidcam-src:
https://unclibrary.atlassian.net/browse/HYC-1678

* Small modifications to ingest service to include article ids for reporting purposes
* DimensionsReportingService, DimensionsReportMailer and html template for email report generation
* Tests for aforementioned classes and preview class for rendered html template

### DIFF
--- a/app/services/tasks/dimensions_reporting_service.rb
+++ b/app/services/tasks/dimensions_reporting_service.rb
@@ -25,7 +25,7 @@ module Tasks
     def extract_publication_info
       publication_info = {successfully_ingested: [], failed_to_ingest: [], marked_for_review: []}
       @ingested_publications[:ingested].map do |publication|
-        publication_item = { title: publication['title'], id: publication['id'], url: "https://cdr.lib.unc.edu/concern/articles/#{publication['article_id']}?locale=en", pdf_attached: publication['pdf_attached'] ? 'Yes' : 'No' }
+        publication_item = { title: publication['title'], id: publication['id'], url: "#{ENV['HYRAX_HOST']}/concern/articles/#{publication['article_id']}?locale=en", pdf_attached: publication['pdf_attached'] ? 'Yes' : 'No' }
         if publication['marked_for_review']
           publication_info[:marked_for_review] << publication_item
         else
--- a/app/services/tasks/dimensions_reporting_service.rb
+++ b/app/services/tasks/dimensions_reporting_service.rb
@@ -1,40 +1,40 @@
 # frozen_string_literal: true
 module Tasks
-    class DimensionsReportingService
-        def initialize(ingested_publications)
-            @ingested_publications = ingested_publications
-        end
-
-        def generate_report
-            report = { successfully_ingested_rows: [], marked_for_review_rows: [], failed_to_ingest_rows: [], subject: [], headers: { }}
-            extracted_info = extract_publication_info()
-            formatted_time = @ingested_publications[:time].strftime("%B %d, %Y at %I:%M %p %Z")
-            report[:subject] = "Dimensions Ingest Report for #{formatted_time}"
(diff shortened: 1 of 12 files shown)


spec/services/tasks/dimensions_reporting_service_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+        expect(info[:url]).to eq("#{ENV['HYRAX_HOST']}/concern/articles/#{sample_array[i]['article_id']}?locale=en") if !failed
+        expect(info[:error]).to eq("StandardError - #{test_err_msg}") if failed
+        if failed
+          expect(info[:pdf_attached]).to be_nil
+        else
+          # Offsetting the index if the sample start index is odd
+          offset_index = sample_start_index.even? ? i : i - 1
+          expect(info[:pdf_attached]).to eq('No') if offset_index.even?
+          expect(info[:pdf_attached]).to eq('Yes') if offset_index.odd?
+        end
```
bbpennel, line 113: I think it'd be clearer to have a single conditional versus a conditional plus inline conditionals
```suggestion
        if failed
          expect(info[:error]).to eq("StandardError - #{test_err_msg}")
          expect(info[:pdf_attached]).to be_nil
        else
          expect(info[:url]).to eq("#{ENV['HYRAX_HOST']}/concern/articles/#{sample_array[i]['article_id']}?locale=en")
          # Offsetting the index if the sample start index is odd
          offset_index = sample_start_index.even? ? i : i - 1
          expect(info[:pdf_attached]).to eq('No') if offset_index.even?
          expect(info[:pdf_attached]).to eq('Yes') if offset_index.odd?
        end
```

app/services/tasks/dimensions_reporting_service.rb
```diff
@@ -0,0 +1,41 @@
...
+    def initialize(ingested_publications)
+      @ingested_publications = ingested_publications
+    end
+
+    def generate_report
+      report = { successfully_ingested_rows: [], marked_for_review_rows: [], failed_to_ingest_rows: [], subject: [], headers: { }}
+      extracted_info = extract_publication_info
+      formatted_time = @ingested_publications[:time].strftime('%B %d, %Y at %I:%M %p %Z')
+      report[:subject] = "Dimensions Ingest Report for #{formatted_time}"
+      report[:headers][:reporting_message] = "Reporting publications from dimensions ingest at #{formatted_time} by #{@ingested_publications[:depositor]}."
```
bbpennel, line 13: I would suggest changing "ingest at" to "ingest on" since "ingest at May 28 at 4:00 PM" reads a little weird.

spec/services/tasks/dimensions_ingest_service_spec.rb
```diff
@@ -137,30 +134,36 @@
...
-        expect(res[:failed].first[:error]).to eq([StandardError.to_s, test_err_msg])
+        expect(actual_failed_publication).to eq(expected_failing_publication)
+        expect(actual_failed_publication_error).to eq([StandardError.to_s, test_err_msg])
+        # Removing article_id from the ingested publications for comparison
+        res[:ingested].each_with_index do |ingested_pub, index|
+          ingested_pub.delete('article_id')
+        end
         expect(res[:ingested]).to match_array(ingested_publications)
         expect(res[:time]).to be_a(Time)
       }.to change { Article.count }.by(ingested_publications.size)
```
bbpennel, line 169: I have a faint memory of this coming up on a previous PR, but I would suggest removing everything from inside this `expect... to change` aside from the line that actually causes the article count to increase, so it is clear what the expectation applies to. I know `res` is needed for the rest of the assertions, but we should be able to keep it in scope

bbpennel (changes requested):

bbpennel (approved):

PR #1103, opened 2024-06-11
HYC-1884 - Dimensions Ingest Rake Task

davidcam-src:
https://unclibrary.atlassian.net/browse/HYC-1884

* Introduces dimensions ingest rake task
* Updates action configuration for dev/test/production environments
* Slight changes made to headers in dimensions query service to get around 400 error codes from dimensions


lib/tasks/dimensions_ingest.rake
```diff
@@ -0,0 +1,49 @@
...
+  task ingest_metadata: :environment do
+    Rake::Task['dimensions:ingest_metadata_task'].invoke
+  end
+
+  desc 'Ingest metadata from Dimensions (implementation)'
+  task ingest_metadata_task: :environment do
+    Rails.logger.info "[#{Time.now}] starting dimensions metadata ingest"
+
+    # Read the last run time from a file
+    file_path = Rails.root.join('log', 'last_dimensions_ingest_run.txt')
```
bbpennel, line 15: I believe that we would want to store this outside of the rails root, since we'll end up with a fresh log directory every time we deploy. I would suggest either something in `ENV['DATA_STORAGE']` or in `/logs/hyc/` (which you could probably get via `ENV['LOGS_PATH`]` and then getting the parent directory).

lib/tasks/dimensions_ingest.rake
```diff
@@ -0,0 +1,49 @@
...
+    last_run_time = File.exist?(file_path) ? Date.parse(File.read(file_path).strip) : nil
+    formatted_last_run_time = last_run_time ? last_run_time.strftime('%Y-%m-%d') : nil
+
+    if last_run_time
+      Rails.logger.info "Last ingest run was at: #{last_run_time}"
+      formatted_last_run_time = last_run_time.strftime('%Y-%m-%d')
```
bbpennel, line 21: this looks the same as line 17

lib/tasks/dimensions_ingest.rake
```diff
@@ -0,0 +1,49 @@
...
+    else
+      Rails.logger.info 'No previous run time found. Starting from default date. (1970-01-01)'
+    end
+    config = {
+      'admin_set' => 'Open_Acess_Articles_and_Book_Chapters',
+      'depositor_onyen' => 'admin'
```
bbpennel, line 27: This might need to be configurable. I don't think there is an `admin` user in prod. We've tended to use rebekah as the depositor onyen for most stuff, which is maybe questionable long term, but prod is restrictive about creating users

lib/tasks/dimensions_ingest.rake
```diff
@@ -0,0 +1,49 @@
...
+    publications = ingest_service.ingest_publications(query_service.query_dimensions(date_inserted: formatted_last_run_time))
+    report = Tasks::DimensionsReportingService.new(publications).generate_report
+    begin
+      DimensionsReportMailer.dimensions_report_email(report).deliver_now
+      Rails.logger.info 'Dimensions ingest report email sent successfully.'
+      Rails.logger.info "Ingested Publications (Total #{publications[:ingested].count}): #{publications[:ingested].map { |pub| pub['id'] }}"
+      Rails.logger.info "Failed Publication Ingest (Total #{publications[:failed].count}): #{publications[:failed].map { |pub| pub['id'] }}"
+      Rails.logger.info "[#{Time.now}] completed dimensions metadata ingest"
+    rescue StandardError => e
+      Rails.logger.error "Failed to send test email: #{e.message}"
```
bbpennel, line 41:
```suggestion
      Rails.logger.error "Failed to send email: #{e.message}"
```

app/services/tasks/dimensions_query_service.rb
```diff
@@ -63,14 +50,51 @@ def query_dimensions(with_doi: true, page_size: 100)
...
         rescue HTTParty::Error, StandardError => e
-          Rails.logger.error("HTTParty error during Dimensions API query: #{e.message}")
-          # Re-raise the error to propagate it up the call stack
+          handle_query_error(e, retries)
+          retries += 1
+          retry if retries <= MAX_RETRIES
           raise e
         end
       end
+      # end
```
bbpennel, line 59: can remove

app/services/tasks/dimensions_query_service.rb
```diff
@@ -63,14 +50,51 @@ def query_dimensions(with_doi: true, page_size: 100)
...
       return all_publications.to_a
     end
 
+    def handle_query_error(error, retries)
```
bbpennel, line 63: Feel free to ignore this one, but I find the mix of retries being handled inside and outside this method a bit confusing. This method could be refocused to `query_needs_retry?(retries)`, and then we'd have

```
# This is up at line 53
Rails.logger.error("HTTParty error during Dimensions API query: #{error.message}")
retry if query_needs_retry?(retries)
raise e

...

def query_needs_retry?(retries)
      retries += 1
      if retries <= MAX_RETRIES
        Rails.logger.warn("Retrying query after #{2**retries} seconds. (Attempt #{retries + 1} of #{MAX_RETRIES})")
        sleep(2**retries) # Using base 2 for exponential backoff
        return true
      end
      false
end
```

(Although in either implementation it might end up logging the error twice when it runs out of retries, so logging of the error could be done conditionally only when there are retries remaining)

bbpennel (changes requested):

lib/tasks/dimensions_ingest.rake
davidcam-src: That sounds reasonable. How would you suggest changing it? I was thinking of introducing an environment variable. Something like:

DIMENSIONS_INGEST_DEPOSITOR_ONYEN = admin (dev, test) rebekah (prod)

app/services/tasks/dimensions_query_service.rb
davidcam-src: Appreciate the suggestion, great idea. I implemented the change but made it so retries is still incremented inside of query_dimensions. For some reason retries doesn't update properly when it's incremented inside of the needs_retry function.

lib/tasks/dimensions_ingest.rake
bbpennel: yeah an environment variable would make sense

app/services/tasks/dimensions_query_service.rb
bbpennel: Apparently ruby passes integer variables by value rather than reference, so if we wanted to be able to increment it in both places we'd have to do something weird like wrap it in an array
```
retries = [0]

def query_needs_retry?(retries)
   retries[0] += 1
   ...
```

Either way, we should make use of the return value from the method in the rescue block, so:
```
        rescue HTTParty::Error, StandardError => e
          retries += 1
          retry if query_needs_retry?(e, retries)
          raise e
        end
```

spec/services/tasks/dimensions_reporting_service_spec.rb
```diff
@@ -59,7 +59,7 @@
     workflow
     workflow_state
     allow(Time).to receive(:now).and_return(fixed_time)
-    allow(User).to receive(:find_by).with(uid: 'admin').and_return(admin)
+    allow(User).to receive(:find_by).with(uid: ENV['DIMENSIONS_INGEST_DEPOSITOR_ONYEN']).and_return(admin)
```
bbpennel, line 62: For tests, particularly when we're defining the value in the test itself, it's fine and probably preferable to keep the actual value in assertions and mocks, since it's a good bit more readable and the environment is consistent

app/services/tasks/dimensions_query_service.rb
davidcam-src: Sorry I thought I had already made the change to make use of the return value. I've made updates to do that as well as wrap 'retries' in an array.

bbpennel (approved):

PR #1104, opened 2024-06-25
HYC 1932 - Update Dimensions Workflow

davidcam-src:
https://unclibrary.atlassian.net/browse/HYC-1932

A bunch of stuff came up during testing, I made some improvements based on our discussions in the above ticket and some other quality of life changes. Including a little more detail than usual since I figure the PR would be difficult to follow otherwise:

* Changed mapping in DimensionsIngestService to:
  - set `N/A` as the default value if a publication doesn't have an abstract
  - map `concepts` from publications to `fields` in new Articles
  - fetch exclusively the first affiliation for authors to map to `other_affiliations` for creators
 * Changes to DimensionsQueryService:
   - Required start and end_date arguments for query_dimensions. When an earliest date isn't provided the default value is provided from the rake task and I assume the function won't be used much outside of that context **(lines 16-22)**
   - Changed some of the formatting at the start of the query dimensions service. Optional arguments weren't working properly before **(lines 15-18)**
   - Added `dimensions_total_count` class attribute to be used for reporting the amount of publications returned from Dimensions in the email reports **(12-13)**
    - Split the code from `deduplicate_publications` into separate functions since it was getting a bit bloated **(133-185)**
    - Change query string to search inbetween two dates and return fields to include `concepts` **(200-202)**
 * Changes to ReportingService:
     - Specify if the job was run automatically or manually **(17)**
     - Include date range and the amount of publications returned from Dimensions **(18,20)**
 * Updated report email template to include a date range and updated header
 * Changes to Dimensions Rake Task:
     - Take `admin_set`, `start_date` and `end_date` as arguments
     - Require that either bother `start_date` and `end_date` are both provided or neither **(14-19)**
     - Conditionally parse start and end dates. Either `start_date, end_date = provided arguments` or `start_date, end_date = previous_run_time, Today` **(21-40)**
  * Updated query response test fixture to consolidate publications with and without DOIs into one file since the ones without DOIs were mostly there to test marked for review functionality that's been removed
  * Updated ingest test fixture to include a publication with multiple authors to test mapping for authors to creators
  * Changes to DimensionsIngestService Tests:
     - `creates an article with metadata` recovers the article added to our test fixture with multiple creators and checks the mapping for all of them
     - check for abstract and keyword mapping **(265-266)**
     - check for default values in case a publication doesn't have concepts or an abstract **(282-290)**
  * Changes to DimensionsReportingService Tests:
      - check that the service describes the job as manually executed if the argument `is_cron_job` is false **(103-108)**


lib/tasks/dimensions_ingest.rake
```diff
@@ -1,48 +1,72 @@
...
     rescue StandardError => e
       Rails.logger.error "Failed to send email: #{e.message}"
     end
 
-    # Write the last run time to a file
-    File.open(Rails.root.join('log', 'last_dimensions_ingest_run.txt'), 'w') do |f|
-      f.puts Time.current
+    # Write the last run time to a file only if this is a cron job
+    if is_cron_job
+      File.open(Rails.root.join('log', 'last_dimensions_ingest_run.txt'), 'w') do |f|
```
bbpennel, line 65: This isn't the same path as used above for addressing the `last_dimensions_ingest_run.txt` file. I'd suggest adding a helper method to the rake task to compute the path.

lib/tasks/dimensions_ingest.rake
```diff
@@ -1,48 +1,72 @@
...
+      # Read the last run time from a file
+      file_path = File.join(ENV['DATA_STORAGE'], 'last_dimensions_ingest_run.txt')
+      last_run_time = File.exist?(file_path) ? Date.parse(File.read(file_path).strip) : nil
+      if last_run_time
+        Rails.logger.info "Last ingest run was at: #{last_run_time}"
+        start_date = last_run_time.strftime('%Y-%m-%d')
+      else
+        Rails.logger.info "No previous run time found. Starting from default date. (#{EARLIEST_DATE})"
+        start_date = EARLIEST_DATE
+      end
```
bbpennel, line 36: just a readability suggestion, but it might be nice to put all this into a helper method like `get_last_run_time` since this is a rather long method and that's a fairly well contained task.

spec/mailers/previews/dimensions_report_mailer_preview.rb
```diff
@@ -1,6 +1,9 @@
 # frozen_string_literal: true
 # Preview all emails at http://localhost:3000/rails/mailers/dimensions_report_mailer
 class DimensionsReportMailerPreview < ActionMailer::Preview
```
bbpennel, line 3: I'm realizing i don't actually know how the Preview class works. Does this have to be executed like a test, or does it execute when you access the URL in the comment at the top of the file?

spec/services/tasks/dimensions_ingest_service_spec.rb
```diff
@@ -241,31 +241,51 @@
...
       expect(article.valid?).to be true
-      expect(article.title).to eq(['Polypharmacy in US Medicare beneficiaries with antineutrophil cytoplasmic antibody vasculitis'])
-      first_creator = article.creators.find { |creator| creator[:index] == ['1'] }
-      expect(first_creator.attributes['name']).to match_array(['Thorpe, Carolyn T'])
-      expect(first_creator.attributes['other_affiliation']).to match_array(['Veterans Affairs Pittsburgh Healthcare System, PA.'])
-      expect(first_creator.attributes['orcid']).to match_array(['https://orcid.org/0000-0002-7662-7497'])
-      expect(article.abstract).to include(/Treatment requirements of antineutrophil cytoplasmic autoantibody vasculitis/)
-      expect(article.date_issued).to eq('2023-07-01')
+      expect(article.title).to eq(['Patient Perspectives on Performance of a Smartphone App for Atrial FibrillationSelf-Management'])
+      all_creators_names = article.creators.map { |creator| creator.attributes['name'] }
```
bbpennel, line 257: We should probably assert how many authors are expected, since this test could still pass if the expected_author_metadata assignment above didn't work correctly and didn't load any authors or there were more authors in the generated article than expected.

bbpennel (changes requested):

app/services/tasks/dimensions_query_service.rb
```diff
@@ -139,36 +153,53 @@ def deduplicate_publications(publications)
...
-        Rails.logger.debug("Removed duplicate publication with DOI: #{pub['doi']} and title: #{pub['title']}")
-      end
+      [publications_with_doi, publications_without_doi]
+    end
 
-      # Removing publications that have a matching PMID, PMCID, or title in Solr
-      new_publications_without_doi = publications_without_doi.reject do |pub|
-        query_string = solr_query_builder(pub)
+    def remove_duplicate_publications(publications, type = :other)
+      new_publications = publications.reject do |pub|
```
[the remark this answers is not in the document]
davidcam-src: I didn't know this was a thing, thanks for mentioning it.

app/services/tasks/dimensions_query_service.rb
```diff
@@ -139,36 +153,53 @@ def deduplicate_publications(publications)
...
+      query_string = publication_data.join(' OR ')
+      return query_string
+    end
+
     # Query with paramaters to retrieve publications related to UNC
-    def generate_query_string(date_inserted, page_size, cursor)
-      search_clauses = ['where type = "article"', "date_inserted >= \"#{date_inserted}\""].join(' and ')
-      return_fields = ['basics', 'extras', 'abstract', 'issn', 'publisher', 'journal_title_raw', 'linkout'].join(' + ')
+    def generate_query_string(start_date, end_date, page_size, cursor)
+      search_clauses = ['where type = "article"', "date >= \"#{start_date}\"", "date <= \"#{end_date}\""].join(' and ')
```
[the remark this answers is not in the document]
davidcam-src: That makes complete sense, I could see how that would be an issue. I just made it so that the end date isn't included.

spec/mailers/previews/dimensions_report_mailer_preview.rb
davidcam-src: I believe your second assumption was correct. Including a quote from ChatGPT I thought could be helpful:

> The Preview class in Rails executes when you access the URL specified in the comment at the top of the file. This URL is a special preview URL generated by Rails that allows you to preview your mailer without sending it. So, when you navigate to this URL in your browser, Rails renders the email preview based on the corresponding methods defined in the Preview class. You don't need to run it like a test; simply accessing the URL triggers the execution and renders the preview

spec/services/tasks/dimensions_ingest_service_spec.rb
bbpennel: Since expected_author_metadata is programmatically populated, if a bug arose that affected the population of this variable (like the way authors were stored in the input changed), we could incorrectly have this test pass since both expected_author_metadata and article.creators could be empty. It's less flexible, but I would suggest hard coding the expected count, which would also make it easier to understand what is expected when reading the code.

lib/tasks/dimensions_ingest.rake
```diff
@@ -62,11 +43,38 @@ namespace :dimensions do
...
       Rails.logger.info 'Not a cron job. Skipping writing last run time to file.'
     end
   end
+
+    # Helper method to compute the last run file path
+  def self.dimensions_last_run_path
+    File.join(ENV['DATA_STORAGE'], 'last_dimensions_ingest_run.txt')
+  end
+
+  def self.get_last_run_time(args)
```
bbpennel, line 59: The scope of this helper is a bit bigger than what I had highlighted originally, which is fine, I'd just suggest reflecting that in the method name, maybe "get_date_range".

Also, while functionally identical, my understanding is that `true` and `false` are the preferred values for assigning boolean variables, while the all caps versions are mostly intended for assigning values to constants, but I don't feel strongly about it.

bbpennel (changes requested):

bbpennel (approved):

PR #1105, opened 2024-07-02
HYC 1935 - Filter Bots out of Download Stats

davidcam-src:
https://unclibrary.atlassian.net/browse/HYC-1935

* Use `browser` gem in `track_downloads` to filter out and log user agents of common bots
* Added test to to verify filtering of certain user agents


spec/controllers/hyrax/downloads_controller_spec.rb
```diff
@@ -87,6 +87,33 @@
         expect(stub).to have_been_requested.times(1) # must be after the method call that creates request
       end
 
+      it 'does not track downloads for any bot user agent' do
+        bot_user_agents.each do |bot_user_agent|
```
bbpennel, line 91: I appreciate the thoroughness of this, and I don't feel strongly that we need to change it, but figured I'd mention that we may want to only test for one or two well known agents instead, since what we're ultimately interested in testing is that our code will catch a bot agent if `browser` thinks the agent is a bot (since this isn't an integration test it might even be appropriate to just mock browser, but i don't think we have to go that far). 

The nice thing about using an external gem is that they can maintain a list of user agents for us and we can rely on it to work as expected thanks to their own tests. Their list will likely drift over time, so we would rather not have our own test start failing if they changed or removed one of the user agents. Anyways, just wanted to explain my reasoning.

app/helpers/hyc_helper.rb
```diff
@@ -1,6 +1,11 @@
 # frozen_string_literal: true
 # Include hyc-specific helper code here instead of in the HyraxHelper to avoid circular dependencies
 module HycHelper
+  def bot_request?(user_agent)
```
bbpennel, line 4: This is only used in one place, do you think it'd make sense to put it in download_analytics_behavior.rb directly until it's needed in other locations, so that it doesn't get unnecessarily included elsewhere?

davidcam-src: Yeah absolutely, that makes sense

bbpennel (approved):

PR #1106, opened 2024-07-09
HYC-1941 - RDF FrozenError

bbpennel:
https://unclibrary.atlassian.net/browse/HYC-1941

* Pull in patch release branch of ruby rdf which may resolve FrozenErrors


davidcam-src (approved):
looks good!

PR #1111, opened 2024-08-02
HYC-1950 - Dimensions Ingest Failed Downloads QA

davidcam-src:
https://unclibrary.atlassian.net/browse/HYC-1950

* URL encoding for Dimensions PDF linkouts
* Retrieval of PDF using the Wiley TDM API for specific publications and error handling for potential rate limits
* Dimensions PDF linkouts included in reports

### DIFF
--- a/app/services/tasks/dimensions_ingest_service.rb
+++ b/app/services/tasks/dimensions_ingest_service.rb
@@ -161,14 +161,13 @@ module Tasks
           # Log a warning if the content type is not a PDF
           raise "Incorrect content type: '#{response.headers['content-type']}'" unless response.headers['content-type']&.include?('application/pdf')
         else
-         # Log a warning if the response code is not 200
+          # Log a warning if the response code is not 200
           Rails.logger.warn("Received a non-200 response code (#{response.code}) when making a HEAD request to the PDF URL: #{encoded_url}")
         end
 
         # Attempt to retrieve the PDF from the encoded URL
         pdf_response = HTTParty.get(encoded_url, headers: headers)
         if pdf_response.code != 200
-          puts "Failed to download PDF: HTTP status '#{pdf_response.code}'"
           e_message = "Failed to download PDF: HTTP status '#{pdf_response.code}'"
           # Include specific error message for potential Wiley-TDM API rate limiting (pdf_response.code != 200 and the Wiley API token was used previously)
           if headers.keys.include?('Wiley-TDM-Client-Token') && wiley_token_used
--- a/app/services/tasks/dimensions_ingest_service.rb
+++ b/app/services/tasks/dimensions_ingest_service.rb
@@ -68,6 +68,7 @@ module Tasks
     def set_basic_attributes(article, publication)
       article.title = [publication['title']]
       article.admin_set = @admin_set
+      article.deposiotr = ENV['DIMENSIONS_INGEST_DEPOSITOR_ONYEN']
       article.creators_attributes = publication['authors'].map.with_index { |author, index| [index, author_to_hash(author, index)] }.to_h
       article.funder = publication['funders']&.map { |funder| funder['name'] }
       article.date_issued = publication['date']
--- a/app/services/tasks/dimensions_ingest_service.rb
+++ b/app/services/tasks/dimensions_ingest_service.rb
@@ -30,7 +30,7 @@ module Tasks
           rescue StandardError => e
             publication.delete('pdf_attached')
             res[:failed] << publication.merge('error' => [e.class.to_s, e.message])
-            Rails.logger.error("Error ingesting publication '#{publication['title']}'")
+            Rails.logger.error("Error ingesting publication '#{publication['title']}' with Dimensions ID: #{publication['id']}")
             Rails.logger.error [e.class.to_s, e.message, *e.backtrace].join($RS)
         end
       end
@@ -41,8 +41,8 @@ module Tasks
       article = article_with_metadata(publication)
       create_sipity_workflow(work: article)
       pdf_path = extract_pdf(publication)
-
       if pdf_path
+        # Rails.logger.info("Attempting to attach PDF to new article for publication with Dimensions ID: #{publication['id']}")
         pdf_file = attach_pdf_to_work(article, pdf_path, @depositor)
         pdf_file.update(permissions_attributes: group_permissions(@admin_set))
         File.delete(pdf_path) if File.exist?(pdf_path)
@@ -131,6 +131,7 @@ module Tasks
     end
 
     def extract_pdf(publication)
+      # Rails.logger.info("Attempting to retrieve PDF for publication with Dimensions ID: #{publication['id']}")
       pdf_url = publication['linkout'] ? publication['linkout'] : nil
       # Set pdf_attached to false by default
       publication['pdf_attached'] = false
@@ -154,6 +155,7 @@ module Tasks
     end
 
     def download_pdf(encoded_url, publication, headers)
+      # Rails.logger.info("Attempting to download PDF for publication with Dimensions ID: #{publication['id']}")
       begin
         # Enforce a delay before making the request
         sleep @download_delay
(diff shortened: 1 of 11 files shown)


app/services/tasks/dimensions_ingest_service.rb
```diff
@@ -130,39 +129,73 @@ def parse_page_numbers(publication)
...
+         # Log a warning if the response code is not 200
+          Rails.logger.warn("Received a non-200 response code (#{response.code}) when making a HEAD request to the PDF URL: #{encoded_url}")
         end
 
-        pdf_response = HTTParty.get(pdf_url)
-        raise "Failed to download PDF: HTTP status '#{pdf_response.code}'" unless pdf_response.code == 200
+        # Attempt to retrieve the PDF from the encoded URL
+        pdf_response = HTTParty.get(encoded_url, headers: headers)
+        if pdf_response.code != 200
+          puts "Failed to download PDF: HTTP status '#{pdf_response.code}'"
```
bbpennel, line 171: Is there a reason to use `puts` here instead of the rails logger?

app/services/tasks/dimensions_ingest_service.rb
```diff
@@ -130,39 +129,73 @@ def parse_page_numbers(publication)
...
-        # Verify the content type is PDF; raise an error if not.
+        # Verify the content type of the PDF before downloading
+        response = HTTParty.head(encoded_url, headers: headers)
         if response.code == 200
+          # Log a warning if the content type is not a PDF
           raise "Incorrect content type: '#{response.headers['content-type']}'" unless response.headers['content-type']&.include?('application/pdf')
         else
-          # Provide a warning and proceed with GET in case the HEAD request failed
-          Rails.logger.warn("Received a non-200 response code (#{response.code}) when making a HEAD request to the PDF URL: #{pdf_url}")
+         # Log a warning if the response code is not 200
```
bbpennel, line 164: indent slightly off

app/services/tasks/dimensions_ingest_service.rb
```diff
@@ -130,39 +129,73 @@ def parse_page_numbers(publication)
...
 
-        pdf_response = HTTParty.get(pdf_url)
-        raise "Failed to download PDF: HTTP status '#{pdf_response.code}'" unless pdf_response.code == 200
+        # Attempt to retrieve the PDF from the encoded URL
+        pdf_response = HTTParty.get(encoded_url, headers: headers)
+        if pdf_response.code != 200
+          puts "Failed to download PDF: HTTP status '#{pdf_response.code}'"
+          e_message = "Failed to download PDF: HTTP status '#{pdf_response.code}'"
+          # Include specific error message for potential Wiley-TDM API rate limiting (pdf_response.code != 200 and the Wiley API token was used previously)
+          if headers.keys.include?('Wiley-TDM-Client-Token') && wiley_token_used
```
bbpennel, line 174: Have you been able to see what one of these rate limited responses looks like? I'm guessing it will have a specific status code, like 429, and these services sometimes include descriptive info in the body of the response which might indicate how long the client is blocked for. If so, we may want to trigger a retry here after some amount of time (like 30 seconds or whatever their response indicates) rather than error out. I needed to do something similar when pulling GA stats into the local hyrax cache, they sent back a long response about the rate limiting.

app/services/tasks/dimensions_ingest_service.rb
```diff
@@ -130,39 +129,73 @@ def parse_page_numbers(publication)
...
+      headers = {}
+
+      unless pdf_url
         Rails.logger.warn('Failed to retrieve PDF. Publication does not have a linkout URL.')
         return nil
       end
+      # Use the Wiley Online Library text data mining API to retrieve the PDF if it's a Wiley publication
+      if pdf_url.include?('hindawi.com') || pdf_url.include?('wiley.com')
+        Rails.logger.info('Detected a Wiley affiliated publication, attempting to retrieve PDF with their API.')
+        @wiley_token_used = true
```
bbpennel, line 144: While it doesn't impact functionality at the moment, I just wanted to mention that by using an instance variable for this it does mean that we wouldn't be able to ingest publications in parallel. It's a good practice to keep the scoping of variables in-line with the whether a service is operating as a singleton, and what the units being processed per method are. In this case, "wiley_token_used" is an attribute of one publication, but its recorded across the whole class which processes multiple publications per instance.

app/views/dimensions_report_mailer/dimensions_report_email.erb
```diff
@@ -27,6 +28,7 @@
           <td style="border: 2px solid #dddddd; text-align: left; padding: 10px;"><%= publication[:id] %></td>
           <td style="border: 2px solid #dddddd; text-align: left; padding: 10px;"><a href="<%= publication[:url] %>"><%= publication[:url] %></a></td>
           <td style="border: 2px solid #dddddd; text-align: left; padding: 10px;"><%= publication[:pdf_attached] %></td>
+          <td style="border: 2px solid #dddddd; text-align: left; padding: 10px;"><%= publication[:linkout] %></td>
```
bbpennel, line 31: I'm guessing this outputs the full URL in a column? It might be better to do something like `<a href="<%= publication[:linkout] %>">Link</a>` in order to keep the table from getting super wide. Might want to add double quote escaping in there just to be safe too.

spec/services/tasks/dimensions_ingest_service_spec.rb
```diff
@@ -230,6 +230,42 @@
...
+        .to_return(body: pdf_content, status: 200, headers: { 'Content-Type' => 'application/pdf' })
+        pdf_path = service.extract_pdf(wiley_test_publication)
+        expect(service).to have_received(:download_pdf).with(
+          "https://api.wiley.com/onlinelibrary/tdm/v1/articles/#{encoded_doi}",
+          wiley_test_publication,
+          'Wiley-TDM-Client-Token' => ENV['WILEY_TDM_API_TOKEN']
+        )
+      end
+
+      it 'raises a unique error if the Wiley API fails and the token has been used previously' do
```
bbpennel, line 261: Its the same token for all requests isn't it? Just not sure if it having been used previously is a factor here.

bbpennel (changes requested):

app/services/tasks/dimensions_ingest_service.rb
```diff
@@ -171,12 +170,23 @@ def download_pdf(encoded_url, publication, headers)
...
         pdf_response = HTTParty.get(encoded_url, headers: headers)
         if pdf_response.code != 200
-          e_message = "Failed to download PDF: HTTP status '#{pdf_response.code}'"
-          # Include specific error message for potential Wiley-TDM API rate limiting (pdf_response.code != 200 and the Wiley API response body mentions rate limiting)
-          if headers.keys.include?('Wiley-TDM-Client-Token') && pdf_response&.body.match?(/rate/i)
-            e_message += ' (Exceeded Wiley-TDM API rate limit)'
+          wiley_rate_exceeded = headers.keys.include?('Wiley-TDM-Client-Token') && pdf_response&.body.match?(/rate/i)
+            # Retry the request after a delay if the Wiley-TDM API rate limit is exceeded
+          if wiley_rate_exceeded
+            delay_time = 30
```
bbpennel, line 176: this is going to add 30 seconds to every build, I'd suggest making this configurable as well (or a factor of the `@download_delay`, like `@download_delay * 15` would get a default 30 seconds)

app/services/tasks/dimensions_ingest_service.rb
```diff
@@ -68,6 +68,7 @@ def populate_article_metadata(article, publication)
     def set_basic_attributes(article, publication)
       article.title = [publication['title']]
       article.admin_set = @admin_set
+      article.depositor = ENV['DIMENSIONS_INGEST_DEPOSITOR_ONYEN']
```
bbpennel, line 71: Other parts of this class are using `@config['depositor_onyen']` to get the depositor_id, I feel like it would be more consistent to use that config entry, or `@depositor.user_key`, just leave dealing with environment variables to rake.

I'm wondering now why the spec files are setting `ENV['DIMENSIONS_INGEST_DEPOSITOR_ONYEN']` since that is only relevant to the `rake` file, which isn't being tested here. I think we could skip the ENV variable management in the tests and just directly set the config value to "admin" for all 4 of the dimensions related tests

bbpennel (changes requested):

bbpennel (approved):

PR #1118, opened 2024-08-30
Hyrax Cache Migration Script QA

davidcam-src:
* Updated work utils helper to:
  * Provide warnings if nothing is found from solr queries instead of raising errors
  * Provide default param in query for admin sets 
* Changed tests to reflect changes
* Moved nil checks outside of utils helper for versatility

### DIFF
--- a/app/helpers/work_utils_helper.rb
+++ b/app/helpers/work_utils_helper.rb
@@ -2,17 +2,20 @@
 module WorkUtilsHelper
   def self.fetch_work_data_by_fileset_id(fileset_id)
     work = ActiveFedora::SolrService.get("file_set_ids_ssim:#{fileset_id}", rows: 1)['response']['docs'].first || {}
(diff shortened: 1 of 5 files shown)


app/helpers/work_utils_helper.rb
```diff
@@ -2,17 +2,30 @@
 module WorkUtilsHelper
   def self.fetch_work_data_by_fileset_id(fileset_id)
     work = ActiveFedora::SolrService.get("file_set_ids_ssim:#{fileset_id}", rows: 1)['response']['docs'].first || {}
-    raise "No work found for fileset id: #{fileset_id}" if work.blank?
+    Rails.logger.warn("No work found for fileset id: #{fileset_id}") if work.blank?
```
bbpennel, line 5: If the work and fileset aren't found, that probably implies that they were deleted. I wonder if we should drop the stats related to it? Might be worth checking with rebekah

bbpennel (approved):

PR #1119, opened 2024-09-23
HYC-1937 - Expand Migration Utility to Include Matomo Stats

davidcam-src:
https://unclibrary.atlassian.net/browse/HYC-1937

* Removed helpers from download analytics behavior, using ENV variables instead
* WorkUtilsHelper now makes separate queries for the work and fileset data 
* ListWorkStatInfo function now calls helpers depending on the data source in the DownloadStatsMigrationService (fetch_matomo_stats and cache_stats)
* Test class for DownloadStatsMigrationService uses a case statement to validate list_work_stat_info and migrate_to_new_table work as expected for each source

### DIFF
--- a/app/services/tasks/download_stats_migration_service.rb
+++ b/app/services/tasks/download_stats_migration_service.rb
@@ -22,6 +22,9 @@ module Tasks
         when DownloadMigrationSource::CACHE
           aggregated_work_stats = fetch_local_cache_stats(after_timestamp, output_path)
           write_to_csv(output_path, aggregated_work_stats)
+        when DownloadMigrationSource::MATOMO
+          aggregated_work_stats = fetch_matomo_stats(after_timestamp, before_timestamp, output_path)
+          write_to_csv(output_path, aggregated_work_stats)
         else
           raise ArgumentError, "Unsupported source: #{source}"
         end
@@ -74,7 +77,7 @@ module Tasks
       Rails.logger.info("Fetching work stats #{timestamp_clause} from Matomo.")
 
       # Query Matomo API for each month in the range and aggregate the data
-      month.each_with_index do |month, index|
+      months_array.each_with_index do |month, index|
         uri_params = {
           module: 'API',
           idSite: matomo_site_id,
--- a/app/services/tasks/download_stats_migration_service.rb
+++ b/app/services/tasks/download_stats_migration_service.rb
@@ -16,38 +16,11 @@ module Tasks
       end
       end
     def list_work_stat_info(output_path, before_timestamp = nil, after_timestamp = nil, source)
+      aggregated_work_stats = []
       begin
         case source
         when DownloadMigrationSource::CACHE
-          query = FileDownloadStat.all
-          query = query.where('updated_at > ?', after_timestamp) if after_timestamp.present?
-          total_work_stats = query.count
-          timestamp_clause = after_timestamp.present? ? "after specified time #{after_timestamp}" : 'without a timestamp'
-
-        # Log number of work stats retrieved and timestamp clause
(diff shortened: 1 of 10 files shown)


app/services/tasks/download_stats_migration_service.rb
```diff
@@ -68,6 +62,88 @@ def migrate_to_new_table(csv_path)
...
+          showColumns: 'nb_events',
+        }
+        reporting_uri.query = URI.encode_www_form(uri_params)
+        response = HTTParty.get(reporting_uri.to_s)
+        Rails.logger.info("Processing Matomo response for #{month}. (#{index + 1}/#{months_array.count})")
+        response.parsed_response.each do |stat|
+          # Events_EventName is the file_id, nb_events is the number of downloads
+          update_aggregate_stats(aggregated_data, month, stat['Events_EventName'], stat['nb_events'])
+        end
+        retrieved_work_stats_total += response.parsed_response.length
```
bbpennel, line 101: So `retrieved_work_stats_total` is a count of the number of days of stats that were found in the time range for this item, is that right? Might help readability to clarify that, took me a little bit to figure out it wasn't the monthly sum of downloads. Maybe something like `days_with_stats_count` or just `days_count`

app/services/tasks/download_stats_migration_service.rb
```diff
@@ -68,6 +62,88 @@ def migrate_to_new_table(csv_path)
...
+    def update_aggregate_stats(aggregated_data, truncated_date, file_id, downloads)
+      # Group the file_id and truncated date to be used as a key
+      key = [file_id, truncated_date]
+      # Initialize the hash for the key if it doesn't exist
+      aggregated_data[key] ||= { file_id: file_id, date: truncated_date, downloads: 0 }
+      # Sum the downloads for each key
+      aggregated_data[key][:downloads] += downloads
+    end
+
+    def first_of_each_month_in_range(after_timestamp, before_timestamp)
```
bbpennel, line 117: just to confirm, the intent here is to capture stats from the beginning of one month through the beginning of another? So if I gave Start: January 22, End: Sept 25, we would get back all the stats from January 1 through Sept 1? Does matomo interpret that to give back all the stats in sept, or just stats for the first day, or no stats from sept?

bbpennel (changes requested):

app/services/tasks/download_stats_migration_service.rb
davidcam-src: I actually intended for `retrieved_work_stats_total` to be a counter for stats returned from Matomo from all dates. 
I thought it would be useful to be able to easily see the ratio of stats returned from them compared to our aggregate. 

Currently stats are retrieved by month so the total number of stats for any given month is represented by `response.parsed_response.length. Would it be fine to instead assign that to a variable `month_stat_count` for clarification? Maybe include some extra comments as well.

bbpennel: sure, that sounds good

app/services/tasks/download_stats_migration_service.rb
davidcam-src: The intent is to actually retrieve stats for every month within the range specified. Since the period `month` is defined in the URI params the Matomo reporting API ignores the specific date and instead just retrieves all stats for that month. 

So for your given example `first_of_each_month_in_range` would truncate the given dates as you said. Then Matomo retrieves the stats for all months specified regardless of date.

bbpennel: Okay, thanks, maybe we should add a short comment to help us remember how matomo works?

davidcam-src: that sounds good

bbpennel (approved):

PR #1123, opened 2024-10-07
HYC-1003 - Text Formatting for Abstracts

davidcam-src:
* Introduced new renderer class so that the Methodology and Abstract fields have rich text formatting


app/assets/javascripts/hyrax/save_work/save_work_control.es6
```diff
@@ -0,0 +1,184 @@
+import { RequiredFields } from './required_fields'
```
bbpennel, line 1: What changed in this file versus the original hyrax version?

app/renderers/hyrax/renderers/formatted_text_renderer.rb
```diff
@@ -0,0 +1,30 @@
...
+
+      # Sanitize the value, allowing only safe HTML tags and attributes
+      def get_sanitized_string(string)
+         # Define allowed tags and attributes
+        allowed_tags = %w[strong em b i u p br small mark sub sup a ul ol li dl dt dd div span h1 h2 h3 h4 h5 h6]
+        allowed_attributes = %w[href]
+        sanitize(string, tags: allowed_tags, attributes: allowed_attributes)
+      end
+
+    # [hyc-override] Same as attribute renderer override, but without escaping the value
```
bbpennel, line 23: This seems to be a new class rather than an override of a hyrax class, so we should remove the `[hyc-override]` part to avoid confusion during upgrades

app/assets/javascripts/hyrax/save_work/required_fields.es6
```diff
@@ -60,4 +60,17 @@ export class RequiredFields {
         this.requiredFields = this.form.find(':input[required], input[name*="date_issued"]');
         this.requiredFields.change(this.callback)
     }
+
+    // [hyc-override] Save TinyMCE content to the textarea before checking for required fields
+    // Important for required fields that are TinyMCE editors
+    saveTinyMCEContent() {
```
bbpennel, line 66: when does this get called? I don't actually see any references to it in this PR, but I might just not understand how it works.

davidcam-src: sorry for the confusion, I rolled back the changes to this file as well as `save_work_control` since they were part of another implementation I was trying out. 

the javascript for saving the tinymce content is in `tinymce.yml` under the rich_text configuration on line 46 if you're interested

bbpennel (approved):

PR #1124, opened 2024-10-08
HYC-1840 - Updates for ruby 3

bbpennel:
https://unclibrary.atlassian.net/browse/HYC-1840

* Updates to using ruby 3, including syntax/language updates to code
* Updates passenger to the second most recent patch version
* Updates to docker base image to add in ruby 3, and work with the current awkward state of centos7 images
* Gems are no longer shared between sidekiq and web docker containers since this wasn't helping much and it didn't work with the necessary extra symlinking steps. Also added a check to see if bundle state was okay before installing

### DIFF
--- a/app/overrides/jobs/attach_files_to_work_job_override.rb
+++ b/app/overrides/jobs/attach_files_to_work_job_override.rb
@@ -6,10 +6,10 @@ Hyrax::AttachFilesToWorkJob.class_eval do
 
   def perform(work, uploaded_files, **work_attributes)
     uploaded_files.each do |uploaded_file|
-      # [hyc-override] check all files for viruses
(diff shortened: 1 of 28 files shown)


app/overrides/jobs/attach_files_to_work_job_override.rb
```diff
@@ -6,10 +6,10 @@
 
   def perform(work, uploaded_files, **work_attributes)
     uploaded_files.each do |uploaded_file|
-      # [hyc-override] check all files for viruses
+      # [hyc-override] check all files for  viruses
```
davidcam-src, line 9:
```suggestion
      # [hyc-override] check all files for viruses
```

minor spacing change

davidcam-src (approved):

PR #1130, opened 2024-11-12
HYC-1938 - Update Migration Utility for GA4 Stats

davidcam-src:


app/services/tasks/download_stats_migration_service.rb
```diff
@@ -62,8 +65,35 @@ def migrate_to_new_table(csv_path)
...
+      Rails.logger.info("Fetching GA4 work stats from specified path #{ga_stats_dir}.")
+
+      # Iterate through each CSV file in the specified directory
+      Dir.glob(File.join(ga_stats_dir, '*.csv')).each do |file_path|
+        range_start_date = retrieve_start_date_from_csv(file_path)
+        unique_row_count = 0
+        Rails.logger.info("Processing file with start date: #{range_start_date}")
+        # Read each CSV file and aggregate data
+        CSV.foreach(file_path, headers: true).with_index do |row, index|
+          next if index < 3  # Skip the first 3 rows
```
bbpennel, line 80: What are the first three rows? It might be nice to mention in the comment

app/services/tasks/download_stats_migration_service.rb
```diff
@@ -145,6 +175,32 @@ def fetch_local_cache_stats(after_timestamp, output_path)
...
+            start_date = DateTime.strptime(date_str, '%Y%m%d')
+            return start_date.strftime('%Y-%m-%d')  # Format to 'YYYY-MM-DD'
+          rescue ArgumentError
+            Rails.logger.error("Invalid date format '#{date_str}' in file #{file_path}")
+            return nil
+          end
+        else
+          Rails.logger.error("Error reading start date from #{file_path}. Second line does not contain a start date.")
+          Rails.logger.error("Found second line: #{second_line}")
+          return nil
```
bbpennel, line 199: Since the application can't really proceed without a start date, I'd suggest having both the cases where the start date can't be determined throw and error rather than return nil

bbpennel (approved):

PR #1132, opened 2024-11-15
HYC-1988 - Handle soffice pptx

bbpennel:
https://unclibrary.atlassian.net/browse/HYC-1988

* If an soffice run times out, the job will attempt a graceful termination, followed by a hard kill. It also now reaps the process if it was killed, to prevent defunct processes from sticking around forever
* Added a lock to soffice processing to prevent multiple runs at once, since apparently it does not support this
    * Added explicit dependency on redlock, which was already getting included via hyrax
* Decreased the number of retries per job from 5 to 3, since jobs rarely recover and block up queues for long periods
* Allow soffice path to be configured from a property

### DIFF
--- a/app/overrides/lib/hydra/derivatives/processors/document_override.rb
+++ b/app/overrides/lib/hydra/derivatives/processors/document_override.rb
@@ -11,38 +11,13 @@ Hydra::Derivatives::Processors::Document.class_eval do
   JOB_TIMEOUT_SECONDS = 5 * 60
   LOCK_MANAGER = Redlock::Client.new([Redis.current])
 
-  # [hyc-override] Adding in a graceful termination before hard kill, use spawn for process group
+  # [hyc-override] Adding in a graceful termination before hard kill, reap the process after kill
   def self.execute_with_timeout(timeout, command, context)
-    stdout, stderr = '', ''
-    pid = nil
-    status = nil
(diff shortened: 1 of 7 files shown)


davidcam-src (approved):

app/overrides/lib/hydra/derivatives/processors/document_override.rb
```diff
@@ -53,28 +28,28 @@ def self.execute_with_timeout(timeout, command, context)
...
+          command = "#{Hydra::Derivatives.libreoffice_path} --invisible --headless --convert-to #{format} --outdir #{outdir} #{Shellwords.escape(path)}"
+          # [hyc-override] Use execute_with_timeout directly
+          execute_with_timeout(timeout, command, {})
+        else
+          sleep(0.5)
+          raise Redlock::LockError, "Failed to acquire lock for document conversion of #{path}"
+        end
       end
+      Rails.logger.debug("Released soffice lock for #{path}")
+    rescue Redlock::LockError
```
davidcam-src, line 50: is it possible that if the lock error is raised with the same path, there might be infinite retries?

davidcam-src: okay cool, thanks

PR #1134, opened 2024-12-05
HYC-1998 - Trigger permission prompt for added embargos

bbpennel:
https://unclibrary.atlassian.net/browse/HYC-1998

* Check if an embargo is added during update of a work, and return that permissions have changed if so

### DIFF
--- a/spec/controllers/concerns/hyrax/works_controller_behavior_spec.rb
+++ b/spec/controllers/concerns/hyrax/works_controller_behavior_spec.rb
@@ -68,8 +68,8 @@ RSpec.describe Hyrax::WorksControllerBehavior, type: :controller do
   describe '#permissions_changed?' do
     let(:user) { FactoryBot.create(:user) }
     let(:work) {
-        General.new(title: ['test work'])
-      }
+      General.new(title: ['test work'])
+    }
 
     context 'with no new permissions or embargo' do
       it 'returns false' do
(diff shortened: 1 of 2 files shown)


spec/controllers/concerns/hyrax/works_controller_behavior_spec.rb
```diff
@@ -61,4 +64,47 @@
...
+
+        controller.send(:save_permissions)
+
+        expect(controller.send(:permissions_changed?)).to be false
+      end
+    end
+
+    context 'with an embargo added' do
+      it 'returns true' do
+        allow(work).to receive(:under_embargo?).and_return(false)
```
davidcam-src, line 87: not sure if this line is intended, noticed similar code on line 92

bbpennel: I was just trying to be explicit that before `save_permissions` was called under_embargo would return false, and then when `permissions_changed` was called under_embargo would return true. This is to simulate the work getting updated with an embargo during the `update` method, which happens in between `save_permissions` and `permissions_changed`

davidcam-src: ah okay gotcha, thanks

davidcam-src (approved):

PR #1136, opened 2024-12-16
HYC-1005 - Skip Ahead in Dept Browse

davidcam-src:
https://unclibrary.atlassian.net/browse/HYC-1005

* Override Blacklight pagination component to enable page skipping similar to catalogue browsing using the Kanamari gem


app/controllers/catalog_controller.rb
```diff
@@ -37,6 +37,45 @@ def single_item_search_builder(id)
...
+    # Facet counts are included with names in the list, so divide by 2
+    total_unique_facets = facet_values ? (facet_values.length / 2) : 0
+    total_unique_facets
+  rescue StandardError => e
+    Rails.logger.error("Error retrieving facets for '#{field_name}': #{e.message}")
+    Rails.logger.error(e.backtrace.join("\n"))
+    0
+  end
+
+  def facet
```
bbpennel, line 65: Is this method called anytime loading a list of facet values? It seems like the same `catalog/facet/` endpoint is getting called both on the departments page and the "more >>" page on a search result. 

My concern is that retrieving a full list of all facet values (which we only display on one page for one facet) could cause slowness for fields other than department. Departmen is limited to a relatively small set of values by its vocab, whereas the creator facet returns 434,000 facet values. Solr is a trooper and only takes 200ms to generate the response, but it takes my browser much longer to download and render the result. So all that to say, maybe we should limit retrieving `total_unique_facets` just when its needed, maybe by checking to see if the facet_field_name is on the list of facet fields that we're allowing to retrieve total counts.

app/components/blacklight/facet_field_pagination_component.html.erb
```diff
@@ -0,0 +1,61 @@
...
+  <ul class="pagination">
+    <% limit = @facet_field.paginator.instance_variable_get(:@limit) %>
+    <% offset = @facet_field.paginator.instance_variable_get(:@offset) %>
+    <% current_page = (offset / limit) + 1 %>
+    <% total_pages = (@total_unique_facets.to_f / limit).ceil %>
+    <% first_pages_count = 2 %> 
+    <% last_pages_count = 2 %> 
+    <% surrounding_pages = 4 %> 
+    <!-- Page Numbers -->
+    <% (1..total_pages).each do |page| %>
```
bbpennel, line 21: Part of me is a little concerned with iterating through the full list of pages, but I think it'll be fine if we limit the pagination to just departments and a select few other fields we trust, as suggested in catalog_controller. 

Just for fun comparison, if you go to a facet browse page on qa for the creator_label_sim field it returns 141 pages. On prod it would return 21700 pages. Iterating 21k times probably still wouldn't have that much performance impact though, I think loading the data would be the main hit.

bbpennel (changes requested):

PR #1143, opened 2025-02-06
HYC-1994 - Viewer Notification Workflow

davidcam-src:
https://unclibrary.atlassian.net/browse/HYC-1994

* Send emails to admin set viewers using DepositedManagerNotification within a new workflow


app/services/hyrax/workflow/deposited_viewer_notification.rb
```diff
@@ -0,0 +1,67 @@
...
+        end
+
+        users_and_roles.each do |query_result|
+          user_id = query_result[:id].to_i
+          user_role_map[user_id] ||= { 'view' => 0, 'manage' => 0 }
+          user_role_map[user_id][query_result[:admin_set_role]] += 1
+        end
+
+        # Select users that have the viewing role applied to them more times than the managing role
+        only_viewers = user_role_map.select { |user_id, role_counts| role_counts['view'] > role_counts['manage'] }
```
bbpennel, line 58: If I'm understanding correctly, if a user is associated with an admin set via multiple assignments (potentially a mix of groups and user roles), this would return true only if the user ends up with more viewer roles than manager? I checked with rebekah, and it sounds like we could actually skip sending viewer notifications if the user has any manager role instead.

Since that's the case, you could potentially have the method return as soon as any manager role is encountered rather than proceeding with both queries, but unless these are slow queries it probably won't have too much of a performance impact.

bbpennel (changes requested):

app/overrides/lib/hyrax/role_registry_override.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+# [hyc-override] https://github.com/samvera/hyrax/blob/hyrax-v4.0.0/lib/hyrax/role_registry.rb
+
+module Hyrax
+  class RoleRegistry
+    MANAGING  = 'managing'
+    APPROVING = 'approving'
+    DEPOSITING = 'depositing'
+    # [hyc-override] Added Viewing role
```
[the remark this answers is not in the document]
davidcam-src: Technically we could avoid adding the viewer role but I think this change results in our update to roles_for_agent being more consistent.

app/services/hyrax/workflow/deposited_viewer_notification.rb
davidcam-src: Yeah that's correct. Got it, I might have been overthinking a bit — made some changes.

bbpennel: After testing out some of the queries, I'm realizing my original suggestions were a bit off base, but the main point of skipping a user with both manager and viewer still remains. I think the new approach may not prevent a user who is directly assigned manager, but is also in a viewer group from getting the notification. 

On cdr-test, if I run both of the queries, I see your onyen returned in both, once with viewer and once with manager. But I don't think the new approach would notice this. It'd produce an array with both entries, and then filter out the one with manager. 

I think the old approach with a minor adjustment would work:
```
        # Map [user_id, group_name, admin_set_role] -> [user_id, {role_name => count}]
        user_role_map = groups_and_roles.each_with_object({}) do |query_result, h|
          user_id = query_result[:user_id].to_i
          h[user_id] ||= { 'view' => 0, 'manage' => 0 }
          h[user_id][query_result[:admin_set_role]] += 1
        end

        users_and_roles.each do |query_result|
          user_id = query_result[:id].to_i
          user_role_map[user_id] ||= { 'view' => 0, 'manage' => 0 }
          user_role_map[user_id][query_result[:admin_set_role]] += 1
        end

        # Filter users to those that have view but no manager
        only_viewers = user_role_map.select { |user_id, role_counts| role_counts['view'] > 0 &&  role_counts['manage'] == 0}  
```

I think we could extend the test https://github.com/UNC-Libraries/hy-c/pull/1143/files#diff-5405747e3eb57393310797274bce9d4aca744a7bddc0ea6fb9b939f8945c8d23R24 to add one more user in that is assigned manager directly and viewer via group1, and verify they don't get an email

bbpennel (approved):

PR #1152, opened 2025-03-19
HYC-2009 - Get tinymce to work on abstract fields after the first one

bbpennel:
https://unclibrary.atlassian.net/browse/HYC-2009

* Overrides FieldManager in order to generate unique IDs for multivalue fields, and handles removing cloned tinymce components and then reinitializing it on new children
*  Assigns unique ids to multivalue fields at page load via new MultiValueWithUniqueIdInput input field type

### DIFF
--- a/app/inputs/multi_value_with_unique_id_input.rb
+++ b/app/inputs/multi_value_with_unique_id_input.rb
@@ -4,7 +4,9 @@ class MultiValueWithUniqueIdInput < MultiValueInput
     options = build_field_options(value, index)
 
     # Generate a unique ID using the field name and index
-    options[:id] = "#{input_dom_id}_#{index}"
+    if index > 0
+      options[:id] = "#{input_dom_id}_#{index}"
+    end
     # Add multi_value class since many parts of the UI rely on it
     options[:class] << 'multi_value'
 
--- a/app/inputs/multi_value_with_unique_id_input.rb
+++ b/app/inputs/multi_value_with_unique_id_input.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+class MultiValueWithUniqueIdInput < MultiValueInput
+  def build_field(value, index)
+    options = build_field_options(value, index)
+
+    # Generate a unique ID using the field name and index
+    options[:id] = "#{input_dom_id}_#{index}"
+    # Add multi_value class since many parts of the UI rely on it
+    options[:class] << 'multi_value'
+
+    if options.delete(:type) == 'textarea'.freeze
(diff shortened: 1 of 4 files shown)


app/inputs/multi_value_with_unique_id_input.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+class MultiValueWithUniqueIdInput < MultiValueInput
+  def build_field(value, index)
+    options = build_field_options(value, index)
```
davidcam-src, line 4: where is `build_field_options` coming from? I'm aware it's a function in `multi_value_select_input` but I thought there'd be another version with an argument override  ([link](https://github.com/samvera/hyrax/blob/cf3f95e9439e4f71de42a8683f7aa868caa1bf5e/app/inputs/multi_value_select_input.rb#L18))

spec/app/inputs/multi_value_with_unique_id_input_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+      end
+
+      it 'generates a text field with ID matching the index' do
+        expect(builder).to receive(:text_field).with(
+          attribute_name,
+          { class: ['multi_value'], id: "#{input_dom_id}_1" }
+        )
+
+        result = input.build_field(value, 1)
+        expect(result).to eq('<input type="text" />')
```
davidcam-src, line 43: does it make sense to also check that the generated field has the correct id?

something like `expected_html = '<input type="text" class="multi_value" id="description_1" />'`

app/inputs/multi_value_with_unique_id_input.rb
bbpennel: It comes from https://github.com/samvera/hydra-editor/blob/96acd9634d0114e667c1773633d2a5f0ca40f85e/app/inputs/multi_value_input.rb#L41
nicely buried in the stack

spec/app/inputs/multi_value_with_unique_id_input_spec.rb
bbpennel: The override doesn't actually set an id when index == 0, so the first instance of the field still has the id "description" rather than "description_0" for backwards compatibility, otherwise a lot of tests break. Normally build_field_options would set id to input_dom_id, but build_field_options is mocked in these tests since its from the parent class. But I think it should be fine for me to have the mock return id=input_dom_id, so I've updated the test to do that, hopefully it'll pass

davidcam-src (approved):

PR #1159, opened 2025-04-03
HYC-2072 - Challenge stats and downloads

bbpennel:


bbpennel:
@lfarrell True. An alternative could be that we don't have any conditional logic in the controllers, always adding the before_action. Then the bot controller would immediately check if turnstile was enabled to centralize it. It'd be the same tiny performance hit (method call to return a cached boolean) as the current approach.

PR #1160, opened 2025-04-04
HYC-2050 - Attach Pubmed Files to Existing Works

davidcam-src:
https://unclibrary.atlassian.net/browse/HYC-2050


lib/tasks/pb_match.rake
```diff
@@ -0,0 +1,159 @@
...
+  res = {skipped: [], successful: [], failed: [], time: Time.now, depositor: DEPOSITOR, directory_or_csv: args[:full_text_dir_or_csv], counts: {}}
+
+#  Retrieve all files within pdf directory
+  file_info =
+  if File.extname(args[:full_text_dir_or_csv]) == '.csv'
+    CSV.read(args[:full_text_dir_or_csv], headers: true)
+    .map { |r| [r['file_name'], r['file_extension']] }
+    .uniq { |file| file[0] } # Deduplicate based on filename
+  else
+    file_info_in_dir(args[:full_text_dir_or_csv])
```
bbpennel, line 35: When would this be used? It seems like we would always be starting from the files listed in the CSV

lib/tasks/pb_match.rake
```diff
@@ -0,0 +1,159 @@
...
+  end
+  # Read the CSV file
+  fetch_identifiers_output_csv = CSV.read(args[:fetch_identifiers_output_csv], headers: true)
```
bbpennel, line 38: I don't totally understand why we are loading the filenames and the CSV separately and then doing a lookup of metadata. Couldn't we just loop through the CSV once since it lists all the files?

bbpennel: This is due to the input files coming from multiple directories, which should be fine at this point

lib/tasks/pb_match.rake
davidcam-src: Yep that's correct, I've removed this

PR #1167, opened 2025-05-06
HYC 2066 - Create new works from PubMed full text

davidcam-src:
https://unclibrary.atlassian.net/browse/HYC-2066

### DIFF
--- a/app/services/tasks/pubmed_ingest_service.rb
+++ b/app/services/tasks/pubmed_ingest_service.rb
@@ -112,10 +112,16 @@ module Tasks
         article.title = metadata.xpath('MedlineCitation/Article/ArticleTitle').text
         article.abstract = metadata.xpath('MedlineCitation/Article/Abstract/AbstractText').text
         article.date_issued = get_date_issued(metadata)
+        publisher = metadata.at_xpath('MedlineCitation/MedlineJournalInfo/MedlineTA')&.text
+        article.publisher = [publisher].compact.presence
+        keywords = metadata.xpath('MedlineCitation/KeywordList/Keyword').map(&:text)
+        article.keyword = keywords if keywords.any?
       elsif metadata.name == 'article'
         article.title = metadata.xpath('front/article-meta/title-group/article-title').text
         article.abstract = metadata.xpath('front/article-meta/abstract').text
         article.date_issued = get_date_issued(metadata)
+        publisher = metadata.at_xpath('front/journal-meta/publisher/publisher-name')&.text
+        article.publisher = [publisher].compact.presence
       else
         # Raise an error for unknown metadata formats
         raise StandardError, "Unknown metadata format: #{metadata.name}"
--- a/app/services/tasks/pubmed_ingest_service.rb
+++ b/app/services/tasks/pubmed_ingest_service.rb
@@ -71,7 +71,7 @@ module Tasks
          skipped_row['pdf_attached'] = 'Success'
          @attachment_results[:successfully_ingested] << skipped_row.to_h
         rescue => e
-          Rails.logger.error("Error ingesting article: #{e.message}")
+          Rails.logger.error(e.message)
           Rails.logger.error(e.backtrace.join("\n"))
           # WIP: Refactoring for error handling, reporting
           skipped_row['pdf_attached'] = e.message
@@ -97,7 +97,7 @@ module Tasks
         identifiers = []
         identifiers << "PMID: #{skipped_row['pmid']}" if skipped_row['pmid'].present?
         identifiers << "PMCID: #{skipped_row['pmcid']}" if skipped_row['pmcid'].present?
-        raise StandardError, "Error during file attachment for new work with the following identifiers: #{identifiers.join(', ')}"
+        raise StandardError, "File attachment error for new work with the following identifiers: #{identifiers.join(', ')}"
       end
       pdf_file.update(permissions_attributes: group_permissions(@admin_set))
       article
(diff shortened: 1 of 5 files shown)


app/services/tasks/pubmed_ingest_service.rb
```diff
@@ -1,16 +1,301 @@
...
+            metadata.at_xpath('PubmedData/ArticleIdList/ArticleId[@IdType="pubmed"]')&.text,
+            metadata.at_xpath('PubmedData/ArticleIdList/ArticleId[@IdType="pmc"]')&.text
+          ]
+        else
+          [
+            metadata.at_xpath('.//article-id[@pub-id-type="pmid"]')&.text,
+            metadata.at_xpath('.//article-id[@pub-id-type="pmcid"]')&.text
+          ]
+        end
+      @new_pubmed_works.find { |row| row['pmid'] == pmid || row['pmcid'] == pmcid }
```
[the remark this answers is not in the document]
davidcam-src: nice catch, updated this

app/services/tasks/pubmed_ingest_service.rb
```diff
@@ -1,16 +1,301 @@
...
+            'index' => i.to_s
+          }
+          # Include affiliations for each author if available
+          retrieve_author_affiliations(res, author, metadata.name, metadata.xpath('front/article-meta/contrib-group'))
+          res
+        end
+      end
+    end
+
+    def retrieve_author_affiliations(hash, author, metadata_name, alt_affiliation_path = nil)
```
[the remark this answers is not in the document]
davidcam-src: working on this atm

app/services/tasks/pubmed_ingest/pubmed_ingest_service.rb
```diff
@@ -0,0 +1,152 @@
...
+        article.rights_statement_label = CdrRightsStatementsService.label(rights_statement)
+        article.dcmi_type = ['http://purl.org/dc/dcmitype/Text']
+      end
+
+      def is_pubmed?(metadata)
+        metadata.name == 'PubmedArticle'
+      end
+
+      def attribute_builder(metadata)
+        is_pubmed?(metadata) ? PubmedAttributeBuilder.new : PmcAttributeBuilder.new
```
bbpennel, line 148: Since this is taking an OOP approach now with a one-builder-per-work structure, I'd suggest storing the metadata object into the builder at construction time rather than passing it into each method. Similarly, it would probably make sense for the builder to hold a ref to the Article object too once it is created. 

It could also make sense to add set_rights_and_types and set_basic_attributes to the builders via a parent class, so that all the attribute population for the works would be internal to the builders. populate_article_metadata could be moved in too so that pubmed_ingest_service would have minimal awareness of the implementation for populating attributes, although that's not always the goal with a builder pattern.

spec/services/tasks/pubmed_ingest/pubmed_ingest_service_spec.rb
```diff
@@ -0,0 +1,450 @@
...
+        service.send(:attach_pdf, article, metadata, skipped_row)
+      }.to raise_error(StandardError, /File attachment error/)
+    end
+  end
+
+  def active_relation_to_string(active_relation)
+    active_relation.to_a.map(&:to_s).join('; ')
+  end
+
+  def build_dynamic_pubmed_xml(count, db_type)
```
bbpennel, line 419: this doesn't seem to be used?

app/services/tasks/pubmed_ingest/base_attribute_builder.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+module Tasks
+  module PubmedIngest
+    # Abstract base class for PubMed and PMC article attribute builders
+    class BaseAttributeBuilder
+      attr_reader :metadata, :article, :admin_set, :depositor_onyen, :new_pubmed_works
+
+      def initialize(metadata, article, admin_set, depositor_onyen, new_pubmed_works)
```
bbpennel, line 8: new_pubmed_works is only used for find_skipped_row, which is directly called by the service and new_pubmed_works isn't really part of the state of the article building built, so I'd suggest passing it in as a param to find_skipped_row (which would also resolve this codeclimate warning).

Alternatively, find_skipped_row could be removed from the builder and replaced by a getter that retrieved the ids (pmid, pcmid, doi). The result could be used in PubmedIngestService to search new_pubmed_works, and could also be used in format_publication_identifiers rather than duplicating the xpath there.

bbpennel (changes requested):

app/services/tasks/pubmed_ingest/pubmed_ingest_service.rb
```diff
@@ -215,10 +217,21 @@ def generate_cdr_url(skipped_row)
...
+        Rails.logger.warn("[generate_cdr_url_for_pubmed_identifier] Failed for identifier: #{identifier}, error: #{e.message}")
         nil
       end
 
+      def generate_cdr_url_for_existing_work(work_id)
+        result = WorkUtilsHelper.fetch_work_data_by_id(work_id)
+        raise "No work found with ID: #{work_id}" if result.nil?
+        raise "Missing work_type for work with id: #{work_id}" unless result[:work_type].present?
+
+        model = result[:work_type]&.underscore&.pluralize || 'works'
```
bbpennel, line 229: given the check at line 227, seems like we don't really need the fallback or &'s here.

It also looks like there isn't any test coverage of this method. I'd be okay if you wanted to test it directly rather than in a bigger test, since the behaviors here seem pretty independent of th rest of the class.

PR #1175, opened 2025-06-26
HYC 1440 - hide video player for large videos

davidcam-src:


bbpennel (approved):
Looks good. I looked at a few of the examples that lead to this ticket, and it seems like the central issue is that there's no webm derivative rather than that the file is too big. But I'm not sure how we'd check that the webm exists, and it doesn't exist because the file is too big for ffmpeg to complete, so this is a good proxy

PR #1233, opened 2026-05-21
Hyc 2167 http longleaf integration

sharonluong:
https://unclibrary.atlassian.net/browse/HYC-2167

### DIFF
--- a/app/jobs/deregister_longleaf_job.rb
+++ b/app/jobs/deregister_longleaf_job.rb
@@ -14,7 +14,7 @@ class DeregisterLongleafJob < Hyrax::ApplicationJob
 
     response = HTTParty.delete(
       "#{base_path}/api/deregister",
-      headers: { "Content-Type": "application/json" },
+      headers: { "Content-Type": 'application/json' },
       body:  { file: path_to_file }.to_json,
       format: :json
     )
--- a/app/jobs/deregister_longleaf_job.rb
+++ b/app/jobs/deregister_longleaf_job.rb
@@ -1,30 +1,40 @@
 # frozen_string_literal: true
-require 'open3'
 
 # Job which causes deregistration of files to longleaf
 class DeregisterLongleafJob < Hyrax::ApplicationJob
   queue_as Hyrax.config.ingest_queue_name
 
   def perform(checksum)
-    if ENV['LONGLEAF_BASE_COMMAND'].blank?
-      Rails.logger.error('LONGLEAF_BASE_COMMAND is not set, skipping deregistration of file to Longleaf.')
-      return
-    end
-
+    base_path = ENV['LONGLEAF_API_HOST_PATH']
     # Calculate the path to the file in fedora, assuming modeshape behavior of hashing based on sha1
-    binary_path = File.join(ENV['LONGLEAF_STORAGE_PATH'], checksum.scan(/.{2}/)[0..2].join('/'), checksum)
-
-    deregister_cmd = "#{ENV["LONGLEAF_BASE_COMMAND"]} deregister -f #{binary_path}"
-
-    Rails.logger.info("Deregistering with longleaf: #{deregister_cmd}")
+    path_to_file = File.join(ENV['LONGLEAF_STORAGE_PATH'], checksum.scan(/.{2}/)[0..2].join('/'), checksum)
 
     start = Time.now
-    stdout, stderr, status = Open3.capture3(deregister_cmd)
+    Rails.logger.debug("Deregistering with longleaf: #{path_to_file}")
 
-    if status.success?
-      Rails.logger.info("Successfully deregistered from Longleaf: #{binary_path}")
+    response = HTTParty.delete(
+      "#{base_path}/api/deregister",
+      headers: { "Content-Type": "application/json" },
+      body:  { file: path_to_file }.to_json,
+      format: :json
+    )
+
+    if response.code == 200
+      body = JSON.parse(response.body)
+      success = body['success']
+      failure = body['failure']
+      unless success.empty?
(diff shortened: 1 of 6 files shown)


app/jobs/deregister_longleaf_job.rb
```diff
@@ -1,30 +1,44 @@
...
+    path_to_file = File.join(ENV['LONGLEAF_STORAGE_PATH'], checksum.scan(/.{2}/)[0..2].join('/'), checksum)
 
     start = Time.now
-    stdout, stderr, status = Open3.capture3(deregister_cmd)
-
-    if status.success?
-      Rails.logger.info("Successfully deregistered from Longleaf: #{binary_path}")
+    Rails.logger.debug("Deregistering with longleaf: #{path_to_file}")
+
+    response = HTTParty.delete(
```
bbpennel: I went back and checked the application, and it looks like its a POST rather than a delete, I guess I wrote the ticket during an earlier stage, sorry!

app/jobs/deregister_longleaf_job.rb
```diff
@@ -1,30 +1,44 @@
...
+
+    if response.code == 200
+      body = JSON.parse(response.body)
+      success = body['success']
+      failure = body['failure']
+      unless success.empty?
+        Rails.logger.info("Successfully deregistered #{path_to_file}")
+      end
+      unless failure.empty?
+        error_message = "Failed to deregister #{path_to_file} from Longleaf. Status code #{response.code}, response body: #{response.body}"
```
bbpennel: I'm trying to decide if displaying the code and body is helpful here. I think the code is guaranteed to be 200 here, and the body will probably not have any other info besides the file path, since this is "successful" failure. On the other hand, are they doing any harm?

app/jobs/deregister_longleaf_job.rb
```diff
@@ -1,30 +1,44 @@
...
+        Rails.logger.error(error_message)
+        raise error_message
+      end
     else
-      Rails.logger.error("Failed to deregister #{binary_path} to Longleaf: #{stdout} #{stderr}")
+      error_message = "Longleaf deregister API returned status #{response.code} for #{path_to_file}"
```
bbpennel: Might be a good idea to include the body here though, just in case the server responded with any kind of info about the error

app/jobs/register_to_longleaf_job.rb
```diff
@@ -1,31 +1,48 @@
...
+      end
+      unless failure.empty?
+        error_message = "Failed to register #{path_to_file} to Longleaf. Status code #{response.code}, response body: #{response.body}"
+        Rails.logger.error(error_message)
+        raise error_message
+      end
     else
-      Rails.logger.error("Failed to register #{binary_path} to Longleaf: #{stdout} #{stderr}")
-      raise "#{stdout} #{stderr}"
+      error_message = "Longleaf register API returned status #{response.code} for #{path_to_file}"
```
bbpennel: Same suggestion about including the body here.

spec/controllers/hyrax/file_sets_controller_spec.rb
```diff
@@ -10,6 +10,14 @@
   describe '#destroy' do
     let(:file_set) { FactoryBot.create(:file_set, :public, :with_original_file, user: user) }
     let(:work) { FactoryBot.create(:work, title: ['test title'], user: user) }
+    let(:longleaf_api_url) { 'https://longleaf.api.com' }
```
bbpennel: i'd suggest using a url like 'https://longleaf.example.com' since that is guaranteed to never resolve to anything, like api.com could resolve to something at some point. I know it's not supposed to be using real clients, but still safer not to risk it

spec/jobs/deregister_longleaf_job_spec.rb
```diff
@@ -2,72 +2,133 @@
...
-      allow(Hyrax::VirusCheckerService).to receive(:file_has_virus?) { false }
-      FileUtils.chmod('u+x', longleaf_script)
-
-      file_set.apply_depositor_metadata admin_user.user_key
-      file_set.save!
-      file_set.original_file = repository_file
-      file_set.save!
+  describe '.perform' do
+    let(:job) { DeregisterLongleafJob.new }
+    let(:longleaf_api_url) { 'https://longleaf.api.com' }
```
bbpennel: same example.com suggestion

spec/jobs/register_to_longleaf_job_spec.rb
```diff
@@ -4,59 +4,143 @@
 require 'tmpdir'
 
 RSpec.describe RegisterToLongleafJob, type: :job do
-  let(:admin_user) { FactoryBot.create(:admin) }
+  describe '.perform' do
+    let(:job) { RegisterToLongleafJob.new }
+    let(:longleaf_api_url) { 'https://longleaf.api.com' }
```
bbpennel: example.com suggestion

spec/models/bulkrax/concerns/file_factory_spec.rb
```diff
@@ -17,6 +17,24 @@
   let(:object) { {} }
   let(:file_set) { FileSet.new }
   let(:temp_pdf_path) { File.join(fixture_path, 'tmp', 'hyrax_test4.pdf') }
+  let(:longleaf_api_url) { 'https://longleaf.api.com' }
```
bbpennel: example.com

bbpennel (changes requested):

app/jobs/deregister_longleaf_job.rb
sharonluong: i removed them!

bbpennel (approved):

PR #1236, opened 2026-05-28
HYC-2162 - Reconcile overrides with upstream changes

bbpennel:
https://unclibrary.atlassian.net/browse/HYC-2167

Reconcile overrides with upstream changes


app/overrides/controllers/hyrax/admin/analytics/work_reports_controller_override.rb
```diff
@@ -1,10 +1,10 @@
 # frozen_string_literal: true
-# https://github.com/samvera/hyrax/blob/hyrax-v4.0.0/app/controllers/hyrax/admin/analytics/work_reports_controller.rb
+# https://github.com/samvera/hyrax/blob/hyrax-v5.2.0/app/controllers/hyrax/admin/analytics/work_reports_controller.rb
 require 'benchmark'
 
 Hyrax::Admin::Analytics::WorkReportsController.class_eval do
   def index
-    return unless Hyrax.config.analytics? && Hyrax.config.analytics_provider != 'ga4'
+    return unless Hyrax.config.analytics_reporting? && Hyrax.config.analytics_provider != 'ga4'
```
sharonluong, line 7: i think it would make more sense if it was `report_analytics?` to evaluate true or false. `analytics_reporting` seems like a noun to me

spec/app/controllers/hyrax/admin/analytics/work_reports_controller_spec.rb
```diff
@@ -9,14 +9,14 @@
     ActiveFedora::Cleaner.clean!
     Blacklight.default_index.connection.delete_by_query('*:*')
     Blacklight.default_index.connection.commit
-    original_analytics = Hyrax.config.analytics?
+    original_analytics = Hyrax.config.analytics_reporting?
```
sharonluong, line 12: not sure what this variable means. does it mean to use the original analytics? trying to figure out what true/false means for `original_analytics`

sharonluong (changes requested):
just some wording change requests, nice work!

app/overrides/controllers/hyrax/admin/analytics/work_reports_controller_override.rb
bbpennel: This is an upstream hyrax thing rather than a local decision unfortunately, so I can't change it

spec/app/controllers/hyrax/admin/analytics/work_reports_controller_spec.rb
bbpennel: its for capturing the original value of `analytics_reporting?` so we can override it for the tests and then restore it back to its original value after the tests

sharonluong (approved):

PR #1238, opened 2026-06-02
HYC-2188 send in one day at a time in pubmed ID query

sharonluong:
https://unclibrary.atlassian.net/browse/HYC-2188

### DIFF
--- a/app/services/tasks/pubmed_ingest/recurring/utilities/id_retrieval_service.rb
+++ b/app/services/tasks/pubmed_ingest/recurring/utilities/id_retrieval_service.rb
@@ -8,18 +8,27 @@ class Tasks::PubmedIngest::Recurring::Utilities::IdRetrievalService
     @end_date = end_date
     @tracker = tracker
   end
+  def retrieve_ids(output_path:, db:)
+    LogUtilsHelper.double_log("Fetching IDs within date range: #{@start_date.strftime('%Y-%m-%d')} -
+      #{@end_date.strftime('%Y-%m-%d')} for #{db} database", :info, tag: 'retrieve_ids')
+    date_ranges = chunkify_date_range(@start_date, @end_date)
+    date_ranges.each do |date_range|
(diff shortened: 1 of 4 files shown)


app/services/tasks/pubmed_ingest/recurring/utilities/id_retrieval_service.rb
```diff
@@ -9,17 +9,28 @@ def initialize(start_date:, end_date:, tracker:)
...
+    dates.each do |date|
+      retrieve_ids_for_one_day(output_path, db, date)
+    end
+  end
+
+  def retrieve_ids_for_one_day(output_path, db, date, retmax: 200, extras: nil)
+    LogUtilsHelper.double_log("Fetching IDs for date: #{date.strftime('%Y-%m-%d')} for #{db} database", :info, tag: 'retrieve_ids_for_one_day')
     base_url = 'https://eutils.ncbi.nlm.nih.gov/entrez/eutils/esearch.fcgi'
     count = 0
     # Initialize cursor from tracker or set to 0
```
bbpennel: This comment isn't accurate anymore, it only sets to 0 now. 

I'm trying to work out what the cursor field in the job_progress is doing at this point. It seems like it is only reporting of the last index reached within a day now, while previously it was supporting resumption. I think losing that is okay, looking at test runs it only takes a couple seconds to pull thousands of ids, and there are checks in `PubmedIngestCoordinatorService` to skip repeating id retrieval if it completed previously. We should probably make the removal of resumption complete though, so here:
https://github.com/UNC-Libraries/hy-c/blob/main/app/services/tasks/pubmed_ingest/recurring/pubmed_ingest_coordinator_service.rb#L116
we would delete record_id_path if it already exists but was not completed previously. Otherwise we could end up with duplicate ids in the file.

bbpennel (changes requested):

bbpennel (approved):

PR #1241, opened 2026-06-10
HYC-2194 update dept vocab list

sharonluong:
https://unclibrary.atlassian.net/browse/HYC-2194

### DIFF
--- a/config/authorities/departments.yml
+++ b/config/authorities/departments.yml
@@ -1400,6 +1400,10 @@ terms:
   term: School of Civic Life and Leadership
   active: true
   short_label: School of Civic Life and Leadership
+- id: School of Data Science and Information Sciences
+  term: School of Data Science and Information Sciences
+  active: true
+  short_label: School of Data Science and Information Sciences
 - id: School of Data Science and Society
   term: School of Data Science and Society
   active: true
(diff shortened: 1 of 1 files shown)


config/authorities/departments.yml
```diff
@@ -1400,6 +1400,10 @@ terms:
   term: School of Civic Life and Leadership
   active: true
   short_label: School of Civic Life and Leadership
+- id: School of Data Science and Information Sciences
```
bbpennel: On the ticket this is `School of Data and Information Sciences` (without the first science), which is correct?

config/authorities/departments.yml
```diff
@@ -1528,6 +1532,10 @@ terms:
   term: UNC Center for the Business of Health
   active: true
   short_label: UNC Center for the Business of Health
+- id: UNC Center for Thriving Communities
```
bbpennel, line 1535: The ticket also says "Please make the following term visible to admins only: UNC Center for Health Equity Research", which I believe means modifying that entry to set `active: false`

sharonluong: whoops, thanks for the reminder

config/authorities/departments.yml
sharonluong: too many sciences =P

bbpennel (approved):

PR #1243, opened 2026-06-22
Show embargo history to admins

lfarrell:
https://unclibrary.atlassian.net/browse/HYC-2198

### DIFF
--- a/app/overrides/presenters/hyrax/file_set_presenter_override.rb
+++ b/app/overrides/presenters/hyrax/file_set_presenter_override.rb
@@ -1,6 +1,15 @@
 # frozen_string_literal: true
 # [hyc-override] https://github.com/samvera/hyrax/blob/hyrax-v4.0.0/app/presenters/hyrax/file_set_presenter.rb
 Hyrax::FileSetPresenter.class_eval do
+  include EmbargoHistoryPresenter
+
+  def embargo_history
+    history = super
+    return history unless history.empty?
+
+    fetch_parent_presenter&.solr_document&.fetch('embargo_history_ssim', []) || []
+  end
+
   def fetch_parent_presenter
     ids = Hyrax::SolrService.query("{!field f=member_ids_ssim}#{id}", fl: Hyrax.config.id_field, rows: 1)
                             .map { |x| x.fetch(Hyrax.config.id_field) }
--- a/app/overrides/presenters/hyrax/file_set_presenter_override.rb
+++ b/app/overrides/presenters/hyrax/file_set_presenter_override.rb
@@ -10,6 +10,7 @@ Hyrax::FileSetPresenter.class_eval do
     fetch_parent_presenter&.solr_document&.fetch('embargo_history_ssim', []) || []
   end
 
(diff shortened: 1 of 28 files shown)


app/overrides/presenters/hyrax/file_set_presenter_override.rb
```diff
@@ -1,6 +1,16 @@
...
+  include EmbargoHistoryPresenter
+
+  def embargo_history
+    history = super
+    return history unless history.empty?
+
+    fetch_parent_presenter&.solr_document&.fetch('embargo_history_ssim', []) || []
+  end
+
+  # Get parent embargo history, as the file set doesn't have this field.
```
bbpennel: I think this comment ended up on the wrong line

app/presenters/concerns/embargo_history_presenter.rb
```diff
@@ -0,0 +1,10 @@
+# app/presenters/concerns/embargo_history_presenter.rb
+# frozen_string_literal: true
+
+module EmbargoHistoryPresenter
+  extend ActiveSupport::Concern
+
+  def embargo_history
+    solr_document['embargo_history_ssim'] || []
```
bbpennel, line 8: just curious, why did this need a custom getter method rather than joining the long line of `delete ... to: :solr_document` entries in each presenter? Partially asking because in hyrax 5 the lists of delegates goes away, so it'd be good to know if we'll need to keep this one or not

app/views/hyrax/embargoes/_embargo_history.html.erb
```diff
@@ -0,0 +1,13 @@
+<%# To render this view pass `embargo_history` (e.g. `object: curation_concern.embargo_history`) %>
+<%# [hyc-override] Add display text for no embargo history %>
```
bbpennel, line 3: Please add a comment that contains a link to the specific version of the file we are overriding from, like
`# [hyc-override] https://github.com/samvera/hyrax/blob/hyrax-v4.0.0/app/presenters/hyrax/work_show_presenter.rb`
these are handy during upgrades

app/views/hyrax/file_sets/show.html.erb
```diff
@@ -0,0 +1,30 @@
+<% provide :page_title, @presenter.page_title %>
```
bbpennel, line 3: Please add a comment that contains a link to the specific version of the file we are overriding

bbpennel (changes requested):

app/presenters/concerns/embargo_history_presenter.rb
lfarrell: There's already one in Hyrax https://github.com/samvera/hyrax/blob/0ffc6aadf1867de9689ab0aff9d72294419b2d7e/app/presenters/hyrax/embargo_presenter.rb#L23, but pulling in the whole presenter seemed to break things.
If the delegate option is going away, maybe this is the better option?

This also is the same pattern archy uses. Arclight doesn't have the delegate concept.

bbpennel (approved):

bbpennel (approved):

PR #1244, opened 2026-06-23
Add flag to include downloads in turnstile checks

lfarrell:
https://unclibrary.atlassian.net/browse/HYC-2204


app/controllers/bot_detect_controller.rb
```diff
@@ -142,10 +142,43 @@ def self.unc_address?(remote_ip_address)
...
+  def self.downloads_controller?(controller)
+    controller.is_a?(Hyrax::DownloadsController)
+  end
+
+  def self.not_thumbnail?(request)
+    request.query_parameters['file'] != 'thumbnail'
+  end
+
+  def self.not_googlebot?(request)
+    !request.user_agent.to_s.downcase.include?('googlebot')
```
bbpennel: It sounds like google scholar likely comes through as GoogleOther, so if we are going to make an exception that'd probably be the one to target (or both). Anyone can set any agent of course though, so if anyone really wants to get access they could set this agent

app/overrides/controllers/hyrax/downloads_controller_override.rb
```diff
@@ -4,6 +4,21 @@
...
+  end
+
+  def self.cf_challenge_downloads_enabled?
+    env_flag_enabled?('CF_TURNSTILE_ENABLED') &&
+      (FlipFlop.challenge_downloads? || env_flag_enabled?('CF_CHALLENGE_DOWNLOADS'))
+  end
+
+  def self.env_flag_enabled?(key)
+    ENV.fetch(key, 'false').casecmp?('true')
+  end
```
bbpennel: Could this be in the BotDetectController? It seems like we'd want to centralize that logic in the controller dedicated to it

bbpennel (approved):

bbpennel (approved):

PR #1253, opened 2026-07-13
Hyrax 5 analytics

lfarrell:
https://unclibrary.atlassian.net/browse/HYC-2213


app/overrides/services/hyrax/analytics/matomo_override.rb
```diff
@@ -1,85 +1,112 @@
...
+  def split_date_range(date_range)
+    date_range.split(',').map { |date| Date.parse(date) }
   end
 end
+
+# Prepend the override module so its methods run before the original Hyrax Matomo class methods.
+# Then re-extend Hyrax::Analytics::Matomo with the patched ClassMethods module so the new methods
+# are available on the module itself.
+Hyrax::Analytics::Matomo::ClassMethods.prepend(HycMatomoMonthlyEventsOverride)
+Hyrax::Analytics::Matomo.extend(Hyrax::Analytics::Matomo::ClassMethods)
```
bbpennel, line 112: not that i mind the new approach, but i was wondering why we switched to doing the override this way rather than the module_eval and alias_methoding?

lfarrell: The `monthly_events_for_id` was never in scope. If it was moved out of the class methods and everything else was left the same then all the helper methods it used were out of scope.

This way is less error prone, at least according to ChatGPT:

**alias_method is fragile here because**:
* you are patching a concern’s class-method module
* the override is being applied through prepend
* super is the correct way to fall back to the original method in a prepended module

PR #1255, opened 2026-07-15
Limit search

bbpennel:


app/controllers/catalog_controller.rb
```diff
@@ -572,4 +583,65 @@ def facet
...
+                         Array.wrap(values)
+                       end
+
+    flattened_values.flatten.compact_blank.any?
+  end
+
+  def integer_param(key)
+    Integer(params[key], 10)
+  rescue ArgumentError, TypeError
+    0
```
cazzerson: I wonder if we should reject invalid (non-int) integer params. It looks like Blacklight allows manipulation in this case, for example:

`/catalog?q=&page=200x`

cazzerson (changes requested):
Looks great, with just one question. Thanks!

app/controllers/catalog_controller.rb
bbpennel: good catch, doing 200x did indeed bypass the restriction. I've updated the PR

cazzerson (approved):

PR #1256, opened 2026-07-16
Limit rows per page parameter directly

bbpennel:
Limit the per_row and rows GET parameters directly rather than setting a global max_per_page, which impacts unexpected queries


cazzerson (approved):
Looks great

PR #1257, opened 2026-07-16
HYC-2211 - Add bootsnap

bbpennel:
https://unclibrary.atlassian.net/browse/HYC-2211

* Add bootsnap and config, make passenger a production dependency, match bundler version on server

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -162,6 +162,8 @@ GEM
     blacklight_range_limit (8.3.0)
       blacklight (>= 7.25.2, < 9)
       deprecation
+    bootsnap (1.24.6)
+      msgpack (~> 1.2)
     bootstrap (4.6.2.1)
       autoprefixer-rails (>= 9.1.0)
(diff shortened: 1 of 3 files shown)


Gemfile.lock
```diff
@@ -1234,4 +1238,4 @@ RUBY VERSION
   ruby 3.3.10p183
 
 BUNDLED WITH
-  4.0.12
+  4.0.8
```
lfarrell, line 1241: Did you mean to lower the bundler version?

lfarrell (changes requested):
L

Gemfile.lock
bbpennel: It matches the version on the servers now. It's not really necessary so i could roll it back if you like, it was leftover from an afternoon of trying to fix hot restarts in docker (modifying tmp/restart.txt) which never worked out, but i figured having all our environments agree on a bundler version wasn't necessarily a bad thing.

PR #1259, opened 2026-07-17
HYC-2214 - X-Sendfile for downloads

bbpennel:
https://unclibrary.atlassian.net/browse/HYC-2214


app/overrides/controllers/hyrax/downloads_controller_override.rb
```diff
@@ -19,6 +20,53 @@ def set_record_admin_set
...
+  def fedora_binary_store_path
+    ENV['FEDORA_BINARY_STORE_PATH'].presence
+  end
+
+  def normalized_sha1_checksum_for(repository_file)
+    checksum = repository_file.checksum&.value.presence || repository_file.digest&.first.to_s.presence
+    return if checksum.blank?
+
+    normalized_checksum = checksum.sub(/\Aurn:sha1:/, '').sub(/\Asha1:/, '')
+    normalized_checksum if normalized_checksum.match?(/\A\h{40}\z/)
```
sharonluong, line 63: does it return blank if the checksum doesn't match the pattern?

config/application.rb
```diff
@@ -18,7 +18,7 @@ class Application < Rails::Application
       env_file = File.join(Rails.root, 'config', ENV['LOCAL_ENV_FILE'] || 'local_env.yml')
 
       YAML.load(File.open(env_file)).each do |key, value|
-        ENV[key.to_s] = value
+        ENV[key.to_s] = value unless ENV.key?(key.to_s)
```
sharonluong, line 21: what is this actually saying? set it to the value unless the key is..?

bbpennel: its telling the application to use existing environment variables when they are present rather than using the value from the local_env.yml file. Without this my docker environment would load values from a local_env.yml file from 2024 rather than from the environment variables supplied via the docker/.env file specified in the docker compose file. I'm not totally sure how the local_env.yml file is getting there, perhaps baked into the base image? It's not the same as the local_env_sample.yml file in the project

sharonluong: ahh it's checking to see if the key is already in the env file, i see

app/overrides/controllers/hyrax/downloads_controller_override.rb
bbpennel: it will return a nil

cazzerson (approved):
My human read says this looks great. The machine asks if the new env var should be added to https://github.com/UNC-Libraries/hy-c/edit/main/app/services/ensure_environment_service.rb

sharonluong (approved):

PR #1263, opened 2026-08-04
HYC-2220 - Improve dashboard/work page performance

bbpennel:
https://unclibrary.atlassian.net/browse/HYC-2220

* Gets works dashboard back to a normal loading time (1-2 seconds), it was taking around 25 seconds to load because it was pulling the full fedora record for every admin set in the repo one by one when the page was requested, in order to populate the facet filter list of admin sets (it still works, it just pulls exclusively from solr now)

### DIFF
--- a/app/overrides/controllers/hyrax/my/works_controller_override.rb
+++ b/app/overrides/controllers/hyrax/my/works_controller_override.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
-# https://github.com/samvera/hyrax/blob/hyrax-v5.2.0/app/controllers/hyrax/my/works_controller.rb
+# [hyc-override] https://github.com/samvera/hyrax/blob/hyrax-v5.2.0/app/controllers/hyrax/my/works_controller.rb
 
 Hyrax::My::WorksController.class_eval do
   private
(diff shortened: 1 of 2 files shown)


app/overrides/controllers/hyrax/my/works_controller_override.rb
```diff
@@ -0,0 +1,38 @@
...
+                            has_model_ssim: Hyrax::ModelRegistry.admin_set_rdf_representations.join(',')
+                          },
+                          type: 'terms'
+                        )
+                        .solr_documents(rows: source_ids.length, fl: 'id,title_tesim')
+                        .map do |doc|
+      [Array(doc['title_tesim']).first.to_s, doc['id']]
+    end
+
+    admin_sets_list.sort do |a, b|
```
sharonluong, line 28: can you explain what this part is doing?

bbpennel: it's sorting the list of admin sets so the "Default" admin set is always first, which is the existing hyrax behavior. Just the block labelled with hyc-override is modified

sharonluong (approved):

PR #1277, opened 2026-08-13
Upgrade to Rails 7

lfarrell:
https://unclibrary.atlassian.net/browse/HYC-2223


app/controllers/omniauth_controller.rb
```diff
@@ -23,4 +23,22 @@ def after_sign_out_path_for(resource_or_scope)
       super
     end
   end
+
+  protected
+
+  # [hyc-override] Overriding Devise's default behavior to redirect to an external URL after sign out.
+  # n Rails 7+, redirect_to requires allow_other_host: true for external hosts
```
bbpennel, line 30: Minor typo here, should be "On Rails 7+".

Also, I don't doubt that this was needed, but its a little confusing since at least on cdr-test when you log out you start on cdr-test and get redirected back to cdr-test, so I'm not sure what the "other_host" is here. In qa and prod where we use shibboleth it makes a bit more sense to me, but I don't remember how this interacts with shibb

app/overrides/controllers/hyrax/downloads_controller_override.rb
```diff
@@ -32,7 +42,7 @@ def show_active_fedora
     return send_fedora_binary_content(local_original_path) if local_original_path.present?
 
     Rails.logger.warn("DownloadsController: falling back to Fedora streaming for FileSet #{params[:id]}")
-    hyc_stream_show_active_fedora
+    super
```
bbpennel, line 45: How is a super call equivalent to calling hyc_stream_show_active_fedora here?

app/overrides/factories/bulkrax/object_factory_override.rb
```diff
@@ -72,16 +95,38 @@ def destroy_unaccounted_for(people_hash, unaccounted_for_ids)
...
+        dcmi_type
+      ]
+
     permitted += %i[admin_set_id] if klass != Collection
+
+    permitted
   end
 end
+
+Bulkrax::ObjectFactory.prepend(HycBulkraxObjectFactoryOverride)
```
bbpennel: why do some classes get guard clauses on the prepends and others don't?

bbpennel:
Here are some comments from the LLM review:

1. Critical config/initializers/bulkrax.rb:17 configures Bulkrax.object_factory as a string rather than the class. Bulkrax 9.5 directly calls both class methods and .new on this value, so CSV validation/import and exports will fail with NoMethodError on "Bulkrax::ObjectFactory". Set it to Bulkrax::ObjectFactory; the local Bulkrax initializer template confirms this expected value.
2. High app/overrides/models/concerns/bulkrax/file_factory_override.rb:24 prepends to Bulkrax::FileFactory, but Bulkrax 9.5 invokes remove_file_set on Bulkrax::FileFactory::InnerWorkings. The override is never called during replacement imports. Replaced files retain Bulkrax’s default removed.png/image/png metadata and are not privatized. The spec calls the method through a direct module include, masking the production dispatch path.
3. High app/overrides/factories/bulkrax/object_factory_override.rb:109 replaces upstream permitted_attributes, omitting Bulkrax 9.5’s newly added embargo/lease fields: visibility_during_embargo, embargo_release_date, visibility_after_embargo, visibility_during_lease, lease_expiration_date, and visibility_after_lease. transform_attributes slices against this list, silently dropping those values from imports.
4. High config/uv/uv.html:53-54 sets the style attribute twice. The second assignment replaces the width declaration, leaving only height. The Universal Viewer container therefore has no explicit width and can render incorrectly or collapse depending on its stylesheet/layout. Use one style assignment or individual style.width and style.height properties.

app/overrides/lib/hydra/services/characterization_service_override.rb
```diff
@@ -0,0 +1,29 @@
...
+# [hyc-override] https://github.com/samvera/hydra-works/blob/v2.3.0/lib/hydra/works/services/characterization_service.rb
+# [hyc-override] This override normalizes scalar and multi-value characterization fields, as some properties (e.g., checksum) are scalar objects,
+# which raises an error when attempting to append values to them as if they were arrays.
+require 'hydra/works/services/characterization_service'
+
+Hydra::Works::CharacterizationService.class_eval do
+  # [hyc-override] Normalize scalar and multi-value characterization fields.
+  # Some properties (e.g., checksum) are scalar objects, so `current + [value]`
+  # would raise when `current` is not an Array.
+  def append_property_value(property, value)
```
bbpennel, line 13: I was wondering what is special about hyc that requires this override. The chatbot thinks its because we are using Wings still, and that it won't be necessary once on valkyrie, but that this problem is actually already present on main. We probably don't need to backport it, but worth keeping in mind that this might not be needed after switching to valkyrie

bin/importer
```diff
@@ -8,7 +8,7 @@ require_relative '../config/environment'
 require 'slop'
 
 def main(opts = {})
-  if opts[:importer_id].blank? || invalid?(opts)
+  if opts[:importer_id].blank? && invalid?(opts)
```
bbpennel: I'm not sure I understand the reason for this change

config/initializers/content_security_policy.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
```
bbpennel, line 1: What is the purpose of this 100% commented out new file?

config/initializers/permissions_policy.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
```
bbpennel, line 1: another fully commented out new file

pubmed.json
```diff
@@ -0,0 +1,24 @@
+{
```
bbpennel: what is this file for? I don't see it referenced anywhere

config/initializers/content_security_policy.rb
lfarrell: It's a new generated config file for Rails 7. It's not strictly necessary to have it, as we just use the defaults.

config/initializers/permissions_policy.rb
lfarrell: A new generated config file for Rails 7. It's not strictly necessary to have it, as we just use the defaults.

pubmed.json
lfarrell: That one snuck in here. That was from some other hy-c thing I was working on a while back. I've deleted it.

app/overrides/factories/bulkrax/object_factory_override.rb
lfarrell: I'll make them consistent. It was just checking if the overrides were already loaded.

app/controllers/omniauth_controller.rb
lfarrell: It was causing the tests to fail. It may not be strictly necessary for the actual functionality.

app/overrides/controllers/hyrax/downloads_controller_override.rb
lfarrell: That was an alias for show_active_fedora, an alias which no longer exists. We use the method directly with our overrides and fall back to the regular method if there isn't a local_original_path. Which if I'm looking at it correctly is what the aliased version was doing.

bbpennel (approved):

PR #1290, opened 2026-08-20
HYC-2224 - Failed to delete trophies warnings

bbpennel:
https://unclibrary.atlassian.net/browse/HYC-2224

Checking for object key existense to avoid key not found errors


bbpennel:
This is not working, just going to suggest fixing it in hyrax

PR #1304, opened 2026-08-31
HYC-2249 - Persistent Solr Connections

bbpennel:
https://unclibrary.atlassian.net/browse/HYC-2249

* Explicitly use PooledSolrRepository in catalog controller so its actually used for requests, rather than being initialized and ignored.
    * Revert blacklight.yml
* Update ActiveFedora::SolrService to use persistent connections, since it accounts for about half the solr requests.

### DIFF
--- a/app/repositories/hyc/pooled_solr_repository.rb
+++ b/app/repositories/hyc/pooled_solr_repository.rb
@@ -11,7 +11,12 @@ module Hyc
     end
 
     def self.shared_connection(connection_config)
-      @shared_connection ||= build_raw_connection(connection_config)
+      # Ensure that each passenger processor gets its own connection rather than attempting to share
+      if @shared_connection.nil? || @shared_connection_pid != Process.pid
+        @shared_connection = build_raw_connection(connection_config)
+        @shared_connection_pid = Process.pid
+      end
+      @shared_connection
     end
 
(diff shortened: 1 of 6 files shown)


app/repositories/hyc/pooled_solr_repository.rb
```diff
@@ -11,7 +11,12 @@ def build_connection
     end
 
     def self.shared_connection(connection_config)
-      @shared_connection ||= build_raw_connection(connection_config)
+      # Ensure that each passenger processor gets its own connection rather than attempting to share
```
sharonluong, line 14: It's confusing to me that we're saying each processor gets its own connection and _not_ share, but the variable is called `shared_connection`. Is there a reason for that?

bbpennel: The connection is shared across all requests in the same passenger process, rather than the default behavior that creates a new connection for every request. But its not shared across all the passenger processes since they are essentially separate instances of hyc running in parallel. So its shared one direction but not another. Since each passenger worker can live for as long as passenger does, it means the solr connection is reused for the same amount of time (if the connection dies for other reasons my understanding is it will try to open a new one)

sharonluong (approved):

PR #1312, opened 2026-09-16
Hyc 2253 add attributes

sharonluong:
https://unclibrary.atlassian.net/browse/HYC-2253


app/models/jats_ingest_work.rb
```diff
@@ -203,16 +203,25 @@ def publication_date_node_set
...
 
+  # Sage has a new schema but unsure of start date
   def electronic_and_physical_publication_date
-    article_metadata.xpath('.//pub-date[@pub-type="epub-ppub"]')
+    old_schema = article_metadata.xpath('.//pub-date[@pub-type="epub-ppub"]')
+    new_schema = article_metadata.xpath('.//pub-date[@publication-format="electronic-print" and @date-type="pub"]')
+    old_schema.present? ? old_schema : new_schema
   end
 
+  # Sage has a new schema but unsure of start date
```
bbpennel: why is this comment in the PR 4 times? i'm not totally sure what it means

app/models/jats_ingest_work.rb
```diff
@@ -203,16 +203,25 @@ def publication_date_node_set
     end
   end
 
+  # Sage has a new schema but unsure of start date
   def electronic_publication_date
-    article_metadata.xpath('.//pub-date[@pub-type="epub"]')
+    old_schema = article_metadata.xpath('.//pub-date[@pub-type="epub"]')
+    new_schema = article_metadata.xpath('.//pub-date[@publication-format="electronic" and @date-type="pub"]')
+    old_schema.present? ? old_schema : new_schema
```
bbpennel: I don't really know if it matters, but I would generally give precedence to the new schema location over the old if for some weird reason a document had both. In practice I assume that would never happen though

app/models/jats_ingest_work.rb
sharonluong: oh, i guess i wanted a note to explain why it's handling both at the same time. but i can remove!

app/models/jats_ingest_work.rb
sharonluong: easy enough to do, i'll update

bbpennel (approved):

PR #1317, opened 2026-09-21
Hyc 2263 nested sage affiliations

sharonluong:
https://unclibrary.atlassian.net/browse/HYC-2263

The problem is that not only were the `aff` elements nested in the `contrib-group` element, the code searching for affiliation IDs was looking for an `xref` element that didn't exist. 
XML with `xref` elements:
xref-elements

XML with nested `aff` elements:
aff_elements


app/models/jats_ingest_work.rb
```diff
@@ -62,18 +62,24 @@ def contributor_to_hash(contributor, index)
   end
 
   def affiliation_map
-    @affiliation_map ||= document.xpath('//aff').map do |affil|
+    affiliations = document.xpath('//aff')
+    @affiliation_map ||= affiliations.map do |affil|
```
bbpennel: I don't think there is any functional change here, but now subsequent calls to affiliation_map will trigger the xpath again rather than immediately returning the cached value. If so, we should probably revert back to the previous structure.

spec/models/jats_ingest_work_spec.rb
```diff
@@ -147,11 +147,36 @@
...
   end
 
-  context 'with an article that has a physical print and new pub date schema' do
+  context 'with an article that has a new pub date schema' do
     let(:xml_file_path) { File.join(fixture_path, 'sage', '10.1177_2192568219888179_new.xml') }
 
     it 'can return metadata from the xml' do
       expect(work.date_of_publication).to eq '2021-01'
     end
   end
```
bbpennel, line 156: We should add a test similar to "can create a properly constructed person object" in this file, using the new affiliation structure in order to ensure that the affiliations get mapped to authors in the expected way.

app/models/jats_ingest_work.rb
```diff
@@ -62,19 +62,25 @@ def contributor_to_hash(contributor, index)
...
       [affil.attributes['id'].value, affiliation_to_s(affil)]
     end.to_h
   end
 
   def affiliation_ids(elem)
-    references = elem.xpath('xref')
-    references.map do |ref|
+    xref_ids = elem.xpath('xref').map do |ref|
```
bbpennel, line 74: This block is essentially the old non-nested method, is that right? My read is in the old method it used the xref tags within the `<contrib>` element to get the rid in order to do a lookup of affiliation elements elsewhere in the document (affiliation_map), while in the new method it goes and finds sibling `<aff>` elements within the same `<contrib-group>` element and associates all of those affiliations with the author(s?) in the same contrib-group. Is that correct? Could you add comments to the two blocks to indicate that they are for differently structured affiliation mappings so we understand why they both exist in the future?

bbpennel (changes requested):

bbpennel (approved):

